PR #107, opened 2021-09-13
Task scheduler

einarmo:
A fairly abstract construct that I nonetheless ended up needing several times in the OPC-UA extractor and decided to add here, as part of an effort to reduce code duplication in OPC-UA. If it doesn't fit I can just add it to the extractor instead, but I think it is sufficiently general that it doesn't hurt to keep it here.

The idea is that it's a sort of producer-consumer scheduler, where a generic list of items are consumed. Each item may need to be run several times, and may produce a new list of child items to consume.

An example of this is exploring a hierarchy, but also reading history (in that case no new nodes are discovered, but this is still a useful construct).

In addition, it contains a system to share capacity. In OPC-UA this is a system of continuation points, but it could also be a cross-source throttler if history involves reading from multiple sources.

This is intended as a very generic base class, it might be helpful to add other base classes later, to handle more specific cases.

### DIFF
--- a/Cognite.Common/BlockingResourceCounter.cs
+++ b/Cognite.Common/BlockingResourceCounter.cs
@@ -12,12 +12,16 @@ namespace Cognite.Extractor.Common
     public interface IResourceCounter
     {
         /// <summary>
-        /// Take <paramref name="count"/> instances of the resource.
-        /// Returns at least 1, if there are zero instances available, blocks until there are.
+        /// Take up to <paramref name="count"/> instances of the resource.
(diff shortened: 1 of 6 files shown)


lprovensi (approved):
The logic looks solid to me. Some methods like OperationScheduler.Run() are quite complex, so I would be worried about testing all the conditions. However, it does not look like the test coverage is worst now, so the test set may be sufficient.
Overall I believe these add nice improvements to the TaskThrottler as well.

PR #109, opened 2021-09-15
Base extractor class

einarmo:
Base class for extractors, to minimize boilerplate code when creating simple extractors.

This is probably just the first step, for live data. Ideally it should also have features designed for history, and some way to enable state-store without too much boilerplate.

This PR also adds the `PeriodicScheduler` class, which simply runs tasks with a scheduled period, which can be interrupted externally. This is something more advanced extractors often end up needing, and it creates a very nice decentralized system for reading form source systems. In this case it is used heavily in the base extractor.

This PR adds a new main example to the README, and also adds that code to a separate (working) example project.

### DIFF
--- a/ExtractorUtils/BaseExtractor.cs
+++ b/ExtractorUtils/BaseExtractor.cs
@@ -7,7 +7,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Cognite.ExtractorUtils
+namespace Cognite.Extractor.Utils
 {
     /// <summary>
     /// Base class for extractors writing timeseries, events or datapoints.
--- a/ExtractorUtils/BaseExtractor.cs
+++ b/ExtractorUtils/BaseExtractor.cs
@@ -1,0 +1,242 @@
+﻿using Cognite.Extensions;
+using Cognite.Extractor.Common;
+using Cognite.Extractor.Utils;
+using CogniteSdk;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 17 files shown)


ExtractorUtils/BaseExtractor.cs
```diff
@@ -0,0 +1,288 @@
...
+        /// </summary>
+        /// <returns></returns>
+        protected abstract Task Start();
+
+
+        /// <summary>
+        /// Called when the extractor is stopping.
+        /// </summary>
+        /// <returns></returns>
+        protected virtual Task OnStop()
```
mathialo, line 104: Nice - I guess this can be utilized to automate reporting extraction pipeline runs?

mathialo (approved):
Very nice!

Only did a skim through of most of it due to the size of the PR, but structure looks nice! A bit different from the python version (such as the splitting of the base and runner class, and the use of dep injection), but I think adapting to conventions in the language is more than worth it, even though it might cause some mismatch between the two frameworks.

ExtractorUtils/BaseExtractor.cs
mathialo: is there a way to catch if it stopped successfully or not in here?

einarmo: Maybe? I'll have to think about it, ideally it shouldn't require anything from the user, and OnStop is intended as a callback for the extractor developer.

einarmo: I don't think it matters if this fails. If they use the runner method it will be caught there, if not they will have to catch it themselves. Any cleanup the extractor library needs to do is done independent of success or failure here.

PR #110, opened 2021-09-17
Installer template

einarmo:
The idea is to make it as simple as possible to create a .NET installer. It isn't simple, so that involves a decent bit of creative configuration. The end result is a JSON configuration file which must be filled in with some seemingly unnecessary information that I couldn't figure out how to extract from the various project files. These are fed automatically into the .wixproj file.

The build script takes "version" and "description", which are embedded into the executable and expected to be unique to each build (In the OPC-UA extractor, which this is based on, version is a tag, and description is the git commit hash + time), location of the MSBuild executable, and path to the config file.

I've tried to make modifying the template simple as well, but dynamic and conditional XML is not very forgiving, so it is a bit of a mess.

The extractor can be added as a windows service, with optional parameters. This does not work that well with how this is done for PI, where there is a separate service and standalone version, but it works well for OPC-UA where these are merged, and once we add systems for doing the same to the utils it will work well for simple extractors as well.


SetupTemplate/SetupTemplate.wixproj
```diff
@@ -0,0 +1,57 @@
...
+  <Target Name="EnsureWixToolsetInstalled" Condition=" '$(WixTargetsImported)' != 'true' ">
+    <Error Text="The WiX Toolset v3.11 (or newer) build tools must be installed to build this project. To download the WiX Toolset, see http://wixtoolset.org/releases/" />
+  </Target>
+  <PropertyGroup>
+    <DefineConstants>target_version=$(target_version);product_name=$(product_name);product_short_name=$(product_short_name);build_dir=buildoutput\;exe_name=$(exe_name);service_args=$(service_args);service=$(service);config_dir=$(config_dir);upgrade_guid=$(upgrade_guid)</DefineConstants>
+  </PropertyGroup>
+  <Target Name="BeforeBuild">
+    <!-- Additional builds can be added here in a similar fashion -->
+	<Exec Command="dotnet publish $(target_project) -c $(target_configuration) -r $(target_runtime) /p:PublishSingleFile=&quot;true&quot; /p:IncludeNativeLibrariesForSelfExtract=&quot;true&quot; /p:InformationalVersion=&quot;$(target_version)&quot; /p:Description=&quot;$(target_desc)&quot; -o buildoutput\" />
```
einarmo, line 55: Making this for python could be as simple as replacing this line with the proper build action.

cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/installer-template/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1037378&projectid=413&pathid=1)

SetupTemplate/Product.wxs
```diff
@@ -0,0 +1,135 @@
...
+    </Directory>
+  </Fragment>
+
+  <Fragment>
+    <ComponentGroup Id="ExecutableComponentGroup">
+      <?if service='true' ?>
+        <Component Id="Primary_exe" Guid="*" Directory="ExecutableFolder">
+          <File Source="$(var.build_dir)\$(var.exe_name)" />
+
+          <?ifdef service_args?>
```
einarmo, line 72: Because compiletime conditionals are tags, they must be added outside of the entire tag you want to switch on. This is _not_ your average ifdef, the contents must be valid XML (which makes sense).

The consequence is that here where I just want to exclude a single parameter (to make it possible to not have service arguments), we have to replicate the entire tag containing the parameter, and all its contents.

ExampleExtractor/ExampleExtractor.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
```
lprovensi, line 5: have you tried `net5.0` with this?

ExampleExtractor/setup-config.json
```diff
@@ -0,0 +1,14 @@
...
+	"target_project": "../ExampleExtractor/ExampleExtractor.csproj",
+	"target_configuration": "Release",
+	"target_runtime": "win-x64",
+	"product_name": "Cognite Example Extractor",
+	"product_short_name": "ExampleExtractor",
+	"exe_name": "ExampleExtractor.exe",
+	"config_dir": "../ExtractorUtils/config",
+	"service": "true",
+	"service_args": "-s",
+	"upgrade_guid": "E86B5018-8786-4BA2-B151-0675EF39A5F6",
```
lprovensi, line 11: ~~where does this guid comes from? any newly generated?~~
Answered in the readme file

README.md
```diff
@@ -196,6 +196,49 @@ using (var provider = services.BuildServiceProvider()) {
...
+
+ - Update the `setup-config.json` file to suit your needs. The default fields are described below. All config fields except for `setup_project` is injected as a build property with msbuild into the `.wixproj` file.
+ - Modify the `InstBanner.bmp` file in `Resources`, adding your extractor name to the blue field.
+ - Modify or replace the `License.rtf` with a license adapted to your project.
+ - (Optional) Add any new additional config files at the bottom of the `Product.wxs` file, by adding new `Component` blocks after the one for `config.example.yml`
+ - (Optional) Add a cognite icon to your extractor by adding `Resources/black32x32.ico` to the folder for your actual extractor, and adding `<ApplicationIcon>black32x32.ico</ApplicationIcon>` to its .csproj file.
+ - Compile the installer in a windows environment with Wix utils and msbuild installed, using something like the following command:
+ 
+`build.ps1 -b Path\To\MSBuild.exe -v 1.0.0 -d "some description" -c Path\To\setup-config.json`
```
lprovensi, line 210: Do you know if this is something that can be done from GitHub actions? I guess if there is an action template with wix tools installed, it should work

SetupTemplate/Product.wxs
```diff
@@ -0,0 +1,135 @@
...
+                        Interactive="no">
+                <util:ServiceConfig FirstFailureActionType="restart"
+                                    SecondFailureActionType="restart"
+                                    ThirdFailureActionType="restart"
+                                    RestartServiceDelayInSeconds="10" />
+                <ServiceDependency Id="Dnscache" />
+                <ServiceConfig DelayedAutoStart="yes" OnInstall="yes" OnReinstall="yes" />
+              </ServiceInstall>
+            <?endif?>
+            <ServiceControl Id="StartService" Stop="both" Remove="uninstall" Name="OpcuaExtractor" Wait="no" />
```
lprovensi, line 113: ExampleExtractor instead of OpcuaExtractor

lprovensi (approved):
Looks really good. I'll try it as soon as I find the time

README.md
einarmo: You'll need wix toolset, yeah. If that can somehow be installed automatically you're good. You don't need wix 4, I've used 3.x, which has been around for some time, so I think there's a decent chance.

SetupTemplate/Product.wxs
einarmo: Yeah, this needs to be short-name. Good catch.

ExampleExtractor/ExampleExtractor.csproj
einarmo: It will work, I'll change it.

PR #112, opened 2021-09-21
Update packages

einarmo:
Remove IdentityComparer, since Identity now contains prover overrides

### DIFF
--- a/Cognite.Extensions/CogniteUtils.cs
+++ b/Cognite.Extensions/CogniteUtils.cs
@@ -251,7 +251,7 @@ namespace Cognite.Extensions
             {
                 throw new ArgumentNullException(nameof(stream));
             }
-            var ret = new Dictionary<Identity, List<Datapoint>>(new IdentityComparer());
+            var ret = new Dictionary<Identity, List<Datapoint>>();
 
             var idSizeBuffer = new byte[sizeof(ushort)];
 
@@ -305,7 +305,7 @@ namespace Cognite.Extensions
                 }
             }
 
-            return ret.ToDictionary(kvp => kvp.Key, kvp => (IEnumerable<Datapoint>)kvp.Value, new IdentityComparer());
+            return ret.ToDictionary(kvp => kvp.Key, kvp => (IEnumerable<Datapoint>)kvp.Value);
         }
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/update-packages/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1037514&projectid=413&pathid=1)

Cognite.Extensions/CogniteUtils.cs
```diff
@@ -543,56 +543,6 @@ public CogniteUtilsException(string message, Exception innerException) : base(me
         }
     }
 
-    /// <summary>
-    /// Comparer for CogniteSdk Identity objects
-    /// </summary>
-    public class IdentityComparer : IEqualityComparer<Identity>
```
mathialo, line 549: Could anything outside the lib using this? Should it be considered breaking to remove?

mathialo (commented):
A lot cleaner! :)

Cognite.Extensions/CogniteUtils.cs
einarmo: Yes. That said, anything outside the lib should be _strongly_ encouraged not to use it, since it is now native. I'd go so far as to say that removing it is a good thing in that users will be alerted to the fact that they should no longer use it.

einarmo: The advantage of a compiled language is that it won't break their code unless they do something crazy.

Also the only place it's really used (afaik) is in the OPC-UA extractor.

einarmo: Also for a little while longer I think we'll be making breaking changes. Hence the "beta".

mathialo: As long as it's clearly labeled as non-stable yet, I think breaking is fine :) 

We should also get all of those things out of the way before we do a 1.x release

mathialo (approved):

PR #114, opened 2021-09-22
Extraction Pipelines

einarmo:
Just a convenient async disposable class that handles reporting success to the pipeline if it has not already been reported.

The idea is that if the extractor exits cleanly, it should report success. If it fails the exception should be caught and reported as an error.


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/ext-pipes/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1037608&projectid=413&pathid=91)

ExtractorUtils/ExtractionRun.cs
```diff
@@ -0,0 +1,128 @@
...
+    public class ExtractionRunConfig
+    {
+        /// <summary>
+        /// ExternalId of extraction pipeline.
+        /// </summary>
+        public string PipelineId { get; set; }
+        /// <summary>
+        /// Frequency of extraction pipeline updates in seconds.
+        /// </summary>
+        public int Frequency { get; set; }
```
mathialo, line 24: I don't think we need to have frequency configurable, it can just be hardcoded to something sensible, like 10 mins.

ExtractorUtils/Configuration/BaseConfig.cs
```diff
@@ -48,6 +48,11 @@ public override void GenerateDefaults()
             if (Logger == null) Logger = new LoggerConfig();
             if (StateStore == null) StateStore = new StateStoreConfig();
         }
+
+        /// <summary>
+        /// Configuration for automatically reporting extraction pipeline runs.
+        /// </summary>
+        public ExtractionRunConfig ExtractionPipeline { get; set; }
```
mathialo, line 55: I was thinking to have this in the cognite section, and not as its own section, so
``` yaml
cognite:
    ...
    extraction-pipeline:
        id: 123
        # or
        external-id: abc
```
How are data sets typically configured for pi/opc ua?

ExtractorUtils/ExtractorRunner.cs
```diff
@@ -79,34 +76,39 @@ public static class ExtractorRunner
...
+                        onCreateExtractor(destination, extractor);
+                    }
+                    var run = provider.GetService<ExtractionRun>();
+                    try
+                    {
+                        await extractor.Start(source.Token).ConfigureAwait(false);
+                    }
+                    catch (TaskCanceledException) when (source.IsCancellationRequested)
+                    {
+                        log.LogWarning("Extractor stopped manually");
```
mathialo, line 103: Where should successful runs be reported?

mathialo (commented):
Very nice!

Just a few comments/questions

ExtractorUtils/ExtractionRun.cs
einarmo: I'm trying to make something generic. I'd rather have it configurable, to be honest. There might be more than one use case, and keeping it configurable does not hurt. I can remove it from the example config if needed.

ExtractorUtils/Configuration/BaseConfig.cs
einarmo: There isn't really a need to have `id` in there, since pipelines require external id.

I still haven't found a good way to handle dataset externalId/internalId, most resources require internal id, but we usually support externalId.

einarmo: I agree it should probably be in the cognite section, there were some issues with the dependency injection, but I'll figure it out.

ExtractorUtils/ExtractorRunner.cs
einarmo: By default they are reported on disposal of the Run object, so at the end of this block.

einarmo: They may also be reported in the `OnStop` method of BaseExtractor, or at any other point. The run object is designed so that you tell it when it should stop, or it stops automatically if disposed before finished.

ExtractorUtils/ExtractionRun.cs
einarmo: I will add a default value, however.

ExtractorUtils/Configuration/BaseConfig.cs
mathialo: Internal ID is true, didn't think of that. I thought I would reuse the new way of configuring data sets in the python utils for consistency, even though using external id is enough for extpipes (you actually need to use external id when reporting runs iirc, so for extpipes the issue is opposite that of data sets). 

This is completely on the side, but for the internalid/externalid issue with datasets - could we do something similar to the python one where we resolve the data set at startup?

einarmo: We could, adding a "DataSet" abstraction could be useful. That said, you still have to reference that on your own when creating assets/timeseries.

einarmo: I'll add it to the list eventually...

ExtractorUtils/ExtractionRun.cs
mathialo: Leaving parameters confugrable when they don't need to be is an anti-pattern though. It will for sure lead to users putting in weird numbers that don't make much sense, such as either leaving it very course and then complain about extraction pipelines not working, or leaving it to fine and filling up the backend DB with bloating data. 

The granularity for the not-seen notifications is expected to be 1h in the backend, so anything finer than 10 mins is not really necessary anyway.

einarmo: I did not know that. That makes this largely useless for OPC-UA and PI and other real-time data applications.

mathialo: Why?

This is mainly intended for major outages, in which 1h is plenty of granularity based on our user research.

einarmo: The issue is that this is _worse_ than the existing solutions. You are asking users to downgrade from grafana which is fast, to this which is extremely slow in comparison. If the data is lost _forever_ on an outage (which may be the case for OPC-UA), then this is just not good enough.

mathialo: No, we are not asking users to downgrade at all. 

The current way most extractor crashes is discovered is either:

 * with the data quality monitoring tool, which still in beta, and often with a granularity of up to 6-12h
 * by manually discovering a gap in data in some time series

this would be far superior to both of these. 

As far as I know there's only one deployment of OPC UA that actually uses grafana/alert manager for automatic error reporting, that's not the default way most do this. And metrics are not going away, this is a supplement. 

The notifications will include webhooks medium term, and will for example be used to automatically open tickets with our and/or the customer's ticketing systems, automatically call incidents, etc. I have been pushing for this granularity to be as low as possible, but both since the severity is meant to be high for these scenarios, false positives is not something we want. All of the users we have interviewed have said 6h is fine. Maybe we will decrease the granularity to 30 or 15 mins in the future, but for now it looks like it's going to be 1h. 

It's also worth noting that this is only for the cases the extractor did not report a failure itself, so in the majority of cases this would not be needed anyway as the crash would have been caught and a failed run would be reported in a `finally` block. Calling it useless is inaccurate.

mathialo (approved):

PR #118, opened 2021-09-24
Add option to automatically gzip datapoints

einarmo:
The default value and common in config.example.yml is based on some research done while writing this for the SDK. Turns out you need a fairly large number of datapoints before it is worthwhile. This feature counts the number of datapoints (not really the number of bytes, we could count lengths of all strings but I'm uncertain of how much that could cost), and decides on whether or not to use gzip based on a configurable limit.

### DIFF
--- a/Cognite.Extensions/TimeSeries/DataPointExtensions.cs
+++ b/Cognite.Extensions/TimeSeries/DataPointExtensions.cs
@@ -8,6 +8,7 @@ using Prometheus;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.IO.Compression;
 using System.Linq;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


Cognite.Extensions/TimeSeries/DataPointExtensions.cs
```diff
@@ -179,10 +183,32 @@ public static DataPointInsertionRequest ToInsertRequest(this IDictionary<Identit
                 var request = points.ToInsertRequest();
                 try
                 {
-                    using (CdfMetrics.Datapoints.WithLabels("create"))
+                    bool useGzip = false;
+                    if (gzipCountLimit >= 0)
```
emelvarol-cognite, line 187: why do we need this check? count will always be bigger than 0, right? 🤔

emelvarol-cognite: I mean checking count > gzipCountLmit should be sufficient?

einarmo: `gzipCountLimit` is the configuration option. I'd like it to be possible to disable the check entirely by letting it be negative.

einarmo: In general, either we throw an error for negative values or we handle it somehow. Disabling the check entirely if the value is invalid works well.

emelvarol-cognite (approved):
LGTM

Cognite.Extensions/TimeSeries/DataPointExtensions.cs
emelvarol-cognite: I see, it is documented 😅

PR #122, opened 2021-09-28
Allow scheduling tasks without running them immediately

einarmo:
This work also uncovered some really problematic stuff related to async testing and xunit. It turns out xunit uses a really strict task scheduler which easily deadlocks if you _happen to have too many tasks blocking_ independent really of ConfigureAwait, etc.

This makes the tests sometimes magically deadlock for no apparent reason, but it shouldn't really happen outside of a custom threading context.

This is not unusual. Xunit devs are known for making decisions that impose arbitrary restrictions on users in order to dictate their worldview, and refusing to provide a reasonable workaround. https://github.com/xunit/xunit/issues/864

### DIFF
--- a/Cognite.Common/PeriodicScheduler.cs
+++ b/Cognite.Common/PeriodicScheduler.cs
@@ -59,13 +59,16 @@ namespace Cognite.Extractor.Common
         /// <param name="name">Name of task, used to refer to it later</param>
         /// <param name="interval">Interval to schedule on</param>
         /// <param name="operation">Function to call on each iteration</param>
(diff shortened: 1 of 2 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/schedule-without-running/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1037942&projectid=413&pathid=105)

PR #123, opened 2021-09-28
Schedule without running

einarmo:
The feature is just an option to start a new periodic task without running it immediately, which is a very common use-case.

The reason why the diff is so large is that this exposed a very ugly issue that I'm still not entirely sure how happens. There seems to be some sort of deadlock when trying to asynchronously for many tasks. It _might_ be isolated to xunit, which is known to do some weird stuff with threading. It might also be related to the WaitAsync method, but attempts to fix it there did not help.

The solution which finally fixed it was manually reporting task completion once the periodic loop terminates, and not wait for the task to complete on its own. Practically there should be no real difference, except that we remove a "layer" of task completions, and force the task to complete so long as the method is allowed to run. In theory the task itself should be cleaned up. If this is indeed caused by a deadlock when awaiting tasks, not awaiting the task should solve the issue.

### DIFF
--- a/Cognite.Common/PeriodicScheduler.cs
+++ b/Cognite.Common/PeriodicScheduler.cs
@@ -10,14 +10,15 @@ namespace Cognite.Extractor.Common
 {
     internal sealed class PeriodicTask : IDisposable
     {
-        public Task Task { get; set; }
+        public Task Task => TCS.Task;
         public Func<CancellationToken, Task> Operation { get; }
         public ManualResetEvent Event { get; } = new ManualResetEvent(false);
         public TimeSpan Interval { get; }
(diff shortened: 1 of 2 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/schedule-without-running/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1037961&projectid=413&pathid=105)

mathialo (approved):

PR #124, opened 2021-09-29
Optionally report success on startup of extractor run

einarmo:
This required some changes to the startup, but the final solution is generally better. Now the extractor itself is responsible for the run, and not the ExtractorRunner Run method. This is more general and useful if users don't want to use the Run method.

### DIFF
--- a/ExtractorUtils/BaseExtractor.cs
+++ b/ExtractorUtils/BaseExtractor.cs
@@ -116,7 +116,10 @@ namespace Cognite.Extractor.Utils
             finally
             {
                 await OnStop().ConfigureAwait(false);
-                await Run.DisposeAsync().ConfigureAwait(false);
+                if (Run != null)
+                {
+                    await Run.DisposeAsync().ConfigureAwait(false);
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/ext-pipes-continuous/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1038022&projectid=413&pathid=1)

ExtractorUtils/BaseExtractor.cs
```diff
@@ -84,17 +97,34 @@ public virtual async Task Start(CancellationToken token)
...
+                {
+                    await Run.Report(ExtPipeRunStatus.failure, true,
+                        $"Error: {ex.Message}\n{ex.StackTrace}", token).ConfigureAwait(false);
+                }
+                throw;
+            }
             finally
             {
                 await OnStop().ConfigureAwait(false);
+                await Run.DisposeAsync().ConfigureAwait(false);
```
wjoel, line 119: Should there be a null-check for Run here as well?

einarmo: Yes. Good catch.

mathialo (approved):

PR #126, opened 2021-10-04
Fix another edge-case in periodic scheduler

einarmo:
If tasks terminated on their own without _newTaskEvent being set, it would not be re-added, which meant that new tasks would not be detected. This solves the issue, and also prevents unnecessary garbage-tasks from piling up in memory.

### DIFF
--- a/Cognite.Common/PeriodicScheduler.cs
+++ b/Cognite.Common/PeriodicScheduler.cs
@@ -228,7 +228,8 @@ namespace Cognite.Extractor.Common
             var tasks = new List<Task>();
             PeriodicTask failedTask = null;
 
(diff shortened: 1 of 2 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/fix-scheduler-again/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1038263&projectid=413&pathid=107)

PR #143, opened 2021-12-09
Move authenticator

einarmo:
Make the authenticator available and usable from just the Cognite.Extensions library, to make it possible to use the utils for just authentication without pulling libraries for config, metrics, state, etc.

Also add documentation for how to do this.


silvavelosa (approved):
LGTM ✔️

PR #144, opened 2021-12-10
Utility method to group by skipped elements

einarmo:
Useful for reporting errors per element, which is a big part of the purpose of the whole error system.

### DIFF
--- a/Cognite.Extensions/CogniteResult.cs
+++ b/Cognite.Extensions/CogniteResult.cs
@@ -211,6 +211,22 @@ namespace Cognite.Extensions
             res.MergeErrors();
             return res;
         }
+
+        /// <summary>
+        /// Return a grouping of skipped error objects with a list of the errors that caused them to be skipped.
+        /// Useful for reporting a list of individual errors per passed resource.
+        /// </summary>
+        /// <returns>Errors grouped by skipped objects</returns>
+        public IEnumerable<(TError Skipped, IEnumerable<CogniteError<TError>> Errors)> ErrorsBySkipped()
+        {
+            if (Errors == null || !Errors.Any()) return Enumerable.Empty<(TError, IEnumerable<CogniteError<TError>>)>();
+
+            return Errors
+                .SelectMany(err => err.Skipped.Select(skipped => (skipped, err)))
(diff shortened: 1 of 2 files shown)


Cognite.Extensions/CogniteResult.cs
```diff
@@ -211,6 +211,22 @@ public static CogniteResult<TError> Merge(params CogniteResult<TError>[]? result
...
+        /// Useful for reporting a list of individual errors per passed resource.
+        /// </summary>
+        /// <returns>Errors grouped by skipped objects</returns>
+        public IEnumerable<(TError Skipped, IEnumerable<CogniteError<TError>> Errors)> ErrorsBySkipped()
+        {
+            if (Errors == null || !Errors.Any()) return Enumerable.Empty<(TError, IEnumerable<CogniteError<TError>>)>();
+
+            return Errors
+                .SelectMany(err => err.Skipped.Select(skipped => (skipped, err)))
+                .GroupBy(pair => pair.skipped)
```
emelvarol-cognite, line 226: of what type is the skipped property here?  you group by `reason type` or `skipped/not skipped`? just curious

ExtractorUtils.Test/unit/CogniteResultTests.cs
```diff
@@ -337,5 +337,45 @@ public void TestThrowOnError()
...
+                {
+                    Skipped = new[] { "s4", "s2" },
+                    Resource = ResourceType.Unit
+                }
+            });
+
+            var groups = result.ErrorsBySkipped();
+            var groupDict = groups.ToDictionary(pair => pair.Skipped, pair => pair.Errors.ToArray());
+
+            Assert.Single(groupDict["s1"]);
```
emelvarol-cognite, line 365: Ok, I get my answer here in the test. by reason

emelvarol-cognite (approved):

Cognite.Extensions/CogniteResult.cs
einarmo: Skipped is `TError`, so a generic type. I group by whatever the default equality comparer for that type is, which in most cases is going to be reference equality, which works well here. The idea is just to get the list of errors for each skipped element.

PR #147, opened 2021-12-15
[CDF-14261] Timeseries updates

einarmo:
Very similar to assets, but somewhat simpler, since the timeseries API is much cleaner. Except for that the semantics are the same, and there is some shared code.

### DIFF
--- a/Cognite.Extensions/TimeSeries/TimeSeriesUpdateResultHandlers.cs
+++ b/Cognite.Extensions/TimeSeries/TimeSeriesUpdateResultHandlers.cs
@@ -12,7 +12,7 @@ namespace Cognite.Extensions
         {
             if (ex.Missing?.Any() ?? false)
             {
-                if (ex.Message.StartsWith("Time series ids not found", StringComparison.InvariantCultureIgnoreCase))
+                if (ex.Message.StartsWith("Time series not found", StringComparison.InvariantCultureIgnoreCase))
                 {
                     err.Type = ErrorType.ItemMissing;
                     err.Resource = ResourceType.Id;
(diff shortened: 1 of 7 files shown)


Cognite.Extensions/TimeSeries/TimeSeriesUpdateResultHandlers.cs
```diff
@@ -0,0 +1,138 @@
...
+                        {
+                            return Identity.Create(longIdVal.Value);
+                        }
+                        else if (dict.TryGetValue("externalId", out var extIdVal)
+                            && extIdVal is MultiValue.String strIdVal
+                            && strIdVal.Value != null)
+                        {
+                            return Identity.Create(strIdVal.Value);
+                        }
+                        return null!;
```
silvavelosa, line 31: maybe it would be better to make `CogniteError.Values` a `IEnumerable<Identity?>?`.... maybe it's too much of a hassle (and in that case we could leave it for later), but I think of `null!` as a smell that should only appear in tests and very exceptional cases.

silvavelosa (approved):
nice work!
Just a small comment on the use of `null!`... if it's too hard to fix right now, feel free to ignore (but it's worth a revisit, to avoid random null reference exceptions later down the road)

Cognite.Extensions/TimeSeries/TimeSeriesUpdateResultHandlers.cs
einarmo: In principle I agree, the reason why I use it here (and everywhere in this PR, I think), is that the nullable analysis in CDF does not properly handle `Where`. The result here can obviously never be null, since we filter them out, but the compiler doesn't understand that. I _think_ this is in the pipeline, but we would be paying a price of performance to work around it ourselves, I think.

einarmo: Making Values nullable is _definitely_ too much of a hassle, and also not really a good solution, since this is exposed to users, and we _don't_ want nulls there at all.

einarmo: I have one idea, I'll see if I can use it.

PR #148, opened 2021-12-16
Improve limiting by items per time unit

einarmo:
Slightly more clever implementation of this functionality, if we schedule 10 items per 1000 seconds, we really want to run one about every 100 seconds, not 10 in the first 10 seconds and 0 in the remaining time. This makes a few small changes to help with this.

### DIFF
--- a/Cognite.Common/TaskThrottler.cs
+++ b/Cognite.Common/TaskThrottler.cs
@@ -309,14 +309,31 @@ namespace Cognite.Extractor.Common
                         }
 
                         if (_lastWaitTime == DateTime.MinValue) _lastWaitTime = DateTime.UtcNow;
+                        try
+                        {
+                            bool timedOut = await CommonUtils
+                                .WaitAsync(_taskCompletionEvent, waitTime, token)
+                                .ConfigureAwait(false);
+                            if (timedOut) _lastWaitTime = DateTime.MinValue;
+                        }
+                        catch (TaskCanceledException)
+                        {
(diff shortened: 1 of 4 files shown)


Cognite.Common/TaskThrottler.cs
```diff
@@ -281,7 +296,23 @@ private async Task Run()
...
+                        TimeSpan waitTime = _timeUnitChunk;
+                        var now = DateTime.UtcNow;
+                        if (_lastWaitTime + _timeUnitChunk > now)
+                        {
+                            waitTime = _timeUnitChunk - (now - _lastWaitTime);
+                        }
+
+                        if (_lastWaitTime == DateTime.MinValue) _lastWaitTime = DateTime.UtcNow;
+
+                        var res = WaitHandle.WaitAny(new[] { token.WaitHandle, _taskCompletionEvent }, waitTime);
```
silvavelosa, line 313: This method blocks the calling thread... would it make any sense to put it inside `await Task.Run()` and send it to other thread?

silvavelosa (approved):
Good call!
just one question on the use of `WaitHandle.WaitAny()`, since it's a blocking operation inside an async method

Cognite.Common/TaskThrottler.cs
einarmo: Good point. I think there might be a way to do this asynchronously properly, using Task.Run isn't quite right here I think.

PR #150, opened 2021-12-17
[CDF-14308] Configuration to ignore SSL-certificate issues

einarmo:
I don't really like adding this, but this is a frequent issue with extractors, and there is no native way to work around it.

It is necessary to do this twice, since the best way to do it is only available on 2.1. Most users will be using a 2.1 compatible version, so this isn't a big problem. On 2.0 they need to call a method explicitly to use this.

### DIFF
--- a/ExtractorUtils/Cognite/DestinationUtils.cs
+++ b/ExtractorUtils/Cognite/DestinationUtils.cs
@@ -10,6 +10,8 @@ using CogniteSdk;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


silvavelosa (commented):
Is this optional? I mean, a client with strict security requirements might prefer to keep certificate validation to their domain controlled policies, instead of a fixed hash from a config file.

einarmo:
> Is this optional? I mean, a client with strict security requirements might prefer to keep certificate validation to their domain controlled policies, instead of a fixed hash from a config file.

That would be the default behavior from .NET itself. This is in cases where for some reason the customer can't/won't properly configure their environment. We typically see these requests from data scientists on tight schedules trying to get something working. Going through customer IT (which is often out-sourced) might add weeks to the project timeline, which can be fatal for the entire project.

PR #152, opened 2021-12-20
Catch errors when retrieving in methods with error handling

einarmo:
We shouldn't really throw exceptions from these, as we expect them to return error objects. We still treat these errors as fatal.


silvavelosa (approved):
LGTM ✔️

PR #159, opened 2022-01-12
Fix race condition in TaskThrottler

einarmo:
We waited for tasks to complete inside the loop, which meant that if cancelation of the loop was requested while it was in the part of the code _after_ that (for example waiting on the mutex), they would never be awaited.

There is no real reason to wait inside the loop, we can just wait outside it, it is effectively the same.

### DIFF
--- a/Cognite.Common/TaskThrottler.cs
+++ b/Cognite.Common/TaskThrottler.cs
@@ -319,8 +319,6 @@ namespace Cognite.Extractor.Common
                         catch (TaskCanceledException)
                         {
                         }
-                        
-
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/improve-throttler/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1039367&projectid=413&pathid=157)

mathialo (approved):

PR #162, opened 2022-01-25
Test utilities project

einarmo:
I find myself rewriting a bunch of test utilities every time I create a new .NET project using the utils. Having a library we can add this stuff to reduces code duplication and makes it more viable to do slightly more complicated stuff.

The library contains a few useful utilities:

 - The console wrapper hack for getting console output from xunit.
 - Utility methods for polling.
 - Utility methods for running with timeout.
 - A serilog logging sink for xunit.


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/test-utilities-project/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1039699&projectid=413&pathid=1)

Cognite.Testing/TestUtils.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <param name="condition">Method to poll</param>
+        /// <param name="seconds">Number of seconds to wait before failing</param>
+        /// <param name="assertion">Method returning a string to include in the thrown exception</param>
+        /// <exception cref="ArgumentNullException">if condition or assertion are null</exception>
+        /// <exception cref="TrueException">
+        /// If <paramref name="condition"/> does not become true within <paramref name="seconds"/>
+        /// </exception>
+        public static async Task WaitForCondition(Func<Task<bool>> condition, int seconds, Func<string> assertion)
+        {
+            if (condition == null) throw new ArgumentNullException(nameof(condition));
```
mathialo, line 26: I'm not sure what exists in C#, but is there something like [lombok](https://projectlombok.org/features/NonNull)? We're using this type of decorators to do null checks in extration pipelines quite a bit and it's very nice, and way shorter than writing it out every time. Worth investigating at least

mathialo (approved):

Cognite.Testing/TestUtils.cs
einarmo: For reference it is default. I.e. the assumption `condition is not null` is present here, but it only generates a warning if the calling project also has enabled nullable analysis.

PR #163, opened 2022-01-25
Fix testing build

einarmo:
Apparently if you include a dependency to xunit then magically your package must set IsPackable.

### DIFF
--- a/Cognite.Testing/Cognite.Testing.csproj
+++ b/Cognite.Testing/Cognite.Testing.csproj
@@ -24,7 +24,8 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.assert" Version="2.4.1" />
+    <PackageReference Include="xunit.core" Version="2.4.1" />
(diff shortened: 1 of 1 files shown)


cognite-cicd:
---|---|---|---|---
[68](https://github.com/cognitedata/dotnet-extractor-utils/blob/fix-testing-lib-build/Cognite.Metrics/MetricsConfig.cs#L68) |Medium|Heap_Inspection|Cognite.Metrics/MetricsConfig.cs|[Checkmarx](https://cognite.checkmarx.net/CxWebClient/ViewerMain.aspx?scanid=1039700&projectid=413&pathid=1)

exyi (approved):
I guess :sweat_smile: 

Why do we need xunit dependency here, though? https://www.nuget.org/packages/xunit.assert/ or https://www.nuget.org/packages/xunit.core/ could be enough.

einarmo:
> I guess 😅
> 
> Why do we need xunit dependency here, though? https://www.nuget.org/packages/xunit.assert/ or https://www.nuget.org/packages/xunit.core/ could be enough.

Great point. That is a better solution here.

PR #165, opened 2022-02-08
Add some utils for logging in test fixtures

einarmo:
This requires being able to replace the test output helper at test startup, so we need some code to make this possible.


silvavelosa (approved):
✔️

PR #167, opened 2022-02-09
Improve flexibility of the extractor runner

einarmo:
Adds the ability to configure services after loading the config file, which is quite common, as well as the ability to specify which config types should be added to the service collection.


silvavelosa (approved):
LGTM ✔️

PR #169, opened 2022-02-18
Command line utility

einarmo:
We've made command line interfaces like this for several extractors. The System.CommandLine library broke the old way of doing it, and it was also quite verbose. This adds a simple attribute-based system for defining flat command line interfaces.

### DIFF
--- a/ExampleExtractor/Program.cs
+++ b/ExampleExtractor/Program.cs
@@ -62,14 +62,14 @@ class Program
         {
             // This can also be invoked directly in main, to not have a CLI.
             await ExtractorRunner.Run<BaseConfig, MyExtractor>(
-                opt.ConfigPath ?? "config.yml",
-                new[] { 1 },
-                "my-extractor",
-                "myextractor/1.0.0",
-                false,
-                true,
-                true,
-                true,
+                configPath: opt.ConfigPath ?? "config.yml",
+                acceptedConfigVersions: new[] { 1 },
+                appId: "my-extractor",
+                userAgent: "myextractor/1.0.0",
+                addStateStore: false,
+                addLogger: true,
+                addMetrics: true,
+                restart: true,
                 CancellationToken.None).ConfigureAwait(false);
         }, binder);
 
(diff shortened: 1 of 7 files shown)


ExampleExtractor/Program.cs
```diff
@@ -37,20 +39,40 @@ protected override async Task Start()
...
+            // This can also be invoked directly in main, to not have a CLI.
+            await ExtractorRunner.Run<BaseConfig, MyExtractor>(
+                opt.ConfigPath ?? "config.yml",
+                new[] { 1 },
+                "my-extractor",
+                "myextractor/1.0.0",
+                false,
+                true,
+                true,
+                true,
```
mathialo, line 72: What is the best practice for stuff like this in C#? A long list of parameters is not very understandable (without an IDE that can tell you what they are). In python you would use kwargs, in java you would use builder methods, is there any C# way of making a list of `false`s and `true`s less cryptic? 🤔

mathialo (commented):
Is there an `argparse` equivalent we could reuse instead of re-implementing it? Or is every existing solution super-verbose or ill-fitted for some other reason?

ExampleExtractor/Program.cs
einarmo: Yeah, in C# you can definitely do this with keywords as well. I've done it that way in OPC-UA, this is just me being lazy.

einarmo:
> Is there an `argparse` equivalent we could reuse instead of re-implementing it? Or is every existing solution super-verbose or ill-fitted for some other reason?

The existing solution is `System.CommandLine` (A native library), and what this is built on. It just doesn't do everything I wanted it to. It is, however, super flexible, which is why building this extension involved so relatively little code. A similar solution like this third party _might_ exist, but System.CommandLine is really new (still in beta), and I doubt there is exactly what I want. 

Technically, the way to do this (create an object containing the parsed command line options) would mean creating a specialized `Binder` for each extractor, then manually adding each field. This just generalizes that approach by using an attribute and reflection instead.

The hard part, i.e. parsing, help, aliases, etc. is all handled by the library.

mathialo (approved):

PR #171, opened 2022-02-21
Update documentation to better cover simple usage

einarmo:


docfx_project/tutorials/minimal-usage.md
```diff
@@ -0,0 +1,135 @@
...
+    appId: "my-app", // Application id passed with HTTP requests to CDF
+    userAgent: "myapp/1.0.0", // User agent for HTTP requests, including application version
+    addStateStore: false, // True if you want to register code for using a litedb state store
+    addLogger: true, // False if you do not want to register a serilog logger
+    addMetrics: false, // True if you want to push or serve metrics to prometheus
+    config: config // Pass a pre-existing config object
+);
+```
+
+## Using just Cognite.Extensions
```
mathialo, line 104: Could we have similar sections on using just upload queues?

einarmo: They are part of the full package, so you can't use them without the full list of dependencies, but a short guide on how to use them given a `destination` object would be nice.

einarmo: Looking at it, there already is one for each queue.

mathialo (approved):

PR #173, opened 2022-03-11
System for logging config objects

einarmo:
Useful for making extractors log their config when starting, but requires a little manual input to make sure we don't log any secrets. This is adapted from some code I wrote for OPC-UA. It is commonly requested to include the config file in the logs, and this makes that possible, while also excluding sensitive options like passwords and secrets.


mathialo (approved):
As long as this is something that's explicitly enabled it should be fine. Otherwise we would risk some users unexpectedly leaking secrets

PR #182, opened 2022-07-13
Remote config

einarmo:
Support for remote configuration files in the .NET extractor utils.

As demonstrated in the ExampleExtractor, this can be quite simple to use: just inject it optionally in the constructor, then pass it to BaseExtractor, and optionally configure it to refresh with a fixed period.

Unlike the python utils, we make no attempts at handling new detected configuration files, beyond triggering an event, which can stop the extractor or make some other extractor specific change.

The _volume_ of code is unfortunately quite large, just because this feature is involved in many parts of the framework.


kjerand (approved):
Looks nice to me

PR #183, opened 2022-07-27
Periodic scheduler bug for dynamic intervals

kjerand:
The periodic scheduler would often run the task many times in a row when using a dynamic timespan such as a cron expression. This is because sometimes the cron expression would give a very small interval.

This change fixes this by adding a dynamic parameter which checks that the interval is larger than a certain limit and it waits before the task is performed instead of afterwards.

### DIFF
--- a/Cognite.Common/PeriodicScheduler.cs
+++ b/Cognite.Common/PeriodicScheduler.cs
@@ -70,6 +70,7 @@ namespace Cognite.Extractor.Common
         private readonly ManualResetEvent _newTaskEvent = new ManualResetEvent(false);
         private readonly object _taskListMutex = new object();
         private readonly Task _internalLoopTask;
+        private readonly int _limit = 15;
 
         /// <summary>
         /// Number of currently active tasks
@@ -98,7 +99,7 @@ namespace Cognite.Extractor.Common
         /// <param name="operation">Function to call on each iteration</param>
         /// <param name="runImmediately">True to execute the periodic task immediately, false to first
         /// wait until triggered by interval or manually</param>
-        /// <param name="dynamic">Whether the interval is dynamic or not</param>
+        /// <param name="dynamic">Whether the interval is dynamic or not, e.g. cron expression</param>
         public void SchedulePeriodicTask(string? name, ITimeSpanProvider interval,
             Func<CancellationToken, Task> operation, bool runImmediately = true, bool dynamic = false)
         {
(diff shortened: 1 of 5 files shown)


Cognite.Common/PeriodicScheduler.cs
```diff
@@ -98,16 +99,17 @@ public PeriodicScheduler(CancellationToken token)
         /// <param name="operation">Function to call on each iteration</param>
         /// <param name="runImmediately">True to execute the periodic task immediately, false to first
         /// wait until triggered by interval or manually</param>
+        /// <param name="dynamic">Whether the interval is dynamic or not, e.g. cron expression</param>
         public void SchedulePeriodicTask(string? name, ITimeSpanProvider interval,
-            Func<CancellationToken, Task> operation, bool runImmediately = true)
```
einarmo, line 102: Ah, no, this isn't pleasant and won't really work. You are expecting users to have internal knowledge of issues in the scheduler.

I think it might be better to see if you can fix this by just modifying `CronTimeSpanWrapper`. See if this is fixed by just adding a small offset to `DateTime.UtcNow` when getting the next occurence, to add a tiny bit of tolerance. I.e. add one second or something. Line 111 in TimeSpanWrapper.cs

einarmo (changes requested):

Cognite.Common/TimeSpanWrapper.cs
```diff
@@ -21,6 +21,10 @@ public class TimeSpanWrapper : ITimeSpanProvider
         private readonly string defaultUnit;
         private readonly TimeSpan defaultValue;
 
+        /// <summary>
+        /// Whether the interval is dynamic or not, e.g. cron expression
+        /// </summary>
+        public bool IsDynamic { get; set; }
```
einarmo, line 27: `override bool IsDynamic => _expression != null;`

Cognite.Common/PeriodicScheduler.cs
```diff
@@ -28,13 +33,17 @@ public class BasicTimeSpanProvider : ITimeSpanProvider
         /// <inheritdoc />
         public TimeSpan Value { get; }
 
+        /// <inheritdoc />
+        public bool IsDynamic { get; set; }
```
einarmo, line 37: `IsDynamic => false;`

einarmo (changes requested):

einarmo (approved):

PR #184, opened 2022-07-27
High availability support

kjerand:
Adding high availability (HA) support in the .NET extractor utils. 

This implementation uses a Raw database that the extractors communicate with to manage which extractors should be in standby and which one should be active. This is done through the RawExtractorManager class, which is used to manage the extractor and to communicate with the shared state. 

To add HA to an extractor you simply need to call the AddHighAvailability() method in the Start() method and add the RawManagerConfig, as seen in the ExampleExtractor.

The general components are located in the ExtractorManager file, while the classes specific to the Raw implementation are located in the RawExtractorManager folder. 

In addition to the implementation there are integration tests and unit tests testing the functionality.


ExtractorUtils.Test/integration/RawManagerIntegrationTest.cs
```diff
@@ -0,0 +1,273 @@
...
+            Task extractor_1 = CreateExtractor(configPath_1, source_1.Token);
+            Task extractor_2 = CreateExtractor(configPath_2, source_2.Token);
+            Task extractor_3 = CreateExtractor(configPath_3, source_3.Token);
+            Task extractor_4 = CreateExtractor(configPath_4, source_4.Token);
+
+            Task cancel_0 = TurnOffAfterDelay(15000, source_1);
+            Task cancel_1 = TurnOffAfterDelay(20000, source_0);
+            Task cancel_2 = TurnOffAfterDelay(30000, source_2);
+            Task cancel_3 = TurnOffAfterDelay(40000, source_4);
+            Task cancel_4 = TurnOffAfterDelay(50000, source_3);
```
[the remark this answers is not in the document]
einarmo: The issue isn't the current build time, really, but every test adds up, and long running tests discourage future testing.

ExtractorUtils.Test/unit/RawManagerTest.cs
```diff
@@ -0,0 +1,493 @@
...
+using Xunit;
+using Xunit.Abstractions;
+using Cognite.Extractor.Utils;
+using Cognite.Extractor.Testing;
+using Cognite.Extractor.Common;
+using CogniteSdk;
+
+namespace ExtractorUtils.Test.Unit
+{
+    public class RawManagerTest
```
einarmo, line 20: These are some really thorough tests, great!

ExtractorUtils/BaseExtractor.cs
```diff
@@ -143,6 +143,37 @@ public virtual async Task Start(CancellationToken token)
...
+        
+        /// <summary>
+        /// Method called to add high availability to an extractor.
+        /// </summary>
+        /// <param name="config">Configuration object</param>
+        /// <param name="interval">Optional update state interval.</param>
+        /// <param name="inactivityThreshold">Optional threshold for extractor being inactive.</param>
+        /// <returns></returns>
+        public async Task AddHighAvailability(
+            RawManagerConfig config,
```
einarmo, line 155: I'm somewhat uncertain about this method name. It seems to me like it does more than just add high availability, it also waits for activity.

I would also really like it if you could generalize this a bit. I.e. don't take `RawManagerConfig`, take `HighAvailabilityConfig`, then construct a manager based on that. Not entirely sure what the best way to do it is, perhaps create a static `CreateHighAvailabilityManager` method somewhere to check which HA type is specified, then create a manager based on that.

ExtractorUtils/Configuration/BaseConfig.cs
```diff
@@ -310,5 +310,26 @@ public class CertificateConfig
         public IEnumerable<string>? AllowList { get; set; }
     }
 
+    /// <summary>
+    /// Config for a RawExtractorManager.
+    /// </summary>
+    public class RawManagerConfig
```
einarmo, line 316: The whole "Manager" name is alright internally, but it does not make sense outside of that specific context. Looking at this I would have no clue what an "Extractor manager" is, certainly my first guess would _not_ be high availability.

ExtractorUtils/RawExtractorManager/RawExtractorManager.cs
```diff
@@ -0,0 +1,237 @@
...
+using CogniteSdk;
+using Cognite.Extractor.Common;
+using Microsoft.Extensions.Logging;
+
+namespace Cognite.Extractor.Utils
+{
+    /// <summary>
+    /// Class used to manage an extractor using a Raw database.
+    /// </summary>
+    public class RawExtractorManager : IExtractorManager
```
einarmo, line 15: Perhaps add `HA` to the name, i.e. `RawHighAvailabilityManager`, and `IHighAvailabilityManager`. This is extractor utils, after all, the `Extractor` bit is fairly implicit.

ExtractorUtils/RawExtractorManager/RawExtractorManager.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            // If there are no active extractors, start the standby extractor with highest priority.
+            if (!activeExtractorResponsive && responsiveStandbyExtractors.Count > 0)
+            {
+                if (responsiveStandbyExtractors.Min() == _config.Index)
+                {
+                    _logger.LogInformation("Extractor is starting.");
+                    return true;
+                }
+            }
```
einarmo, line 124: Minor thing, but this whole method can be done elegantly with `LINQ`:
```
// You should really make sure to fetch time once before doing a check, for consistency.
var now = DateTime.UtcNow;
var responsiveStandByExtractors = _state.CurrentState
    .Where(extractor => now.Subtract(extractor.TimeStamp) < _inactivityTreshold)
    .Select(extractor => extractor.Index);
if (responseStandByExtractors.Any() && responsiveStandByExtractors.Min() == _config.Index)
{
    ...
}
```

einarmo (changes requested):
Great work, and you are pretty close, but there are some various issues I'd like you to look at.

einarmo (approved):
Great work! Looks good.

PR #188, opened 2022-08-09
HA support using Redis

kjerand:
The integration test will not work because it currently uses localhost instead of a test connection string.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -10,6 +10,10 @@ jobs:
   build:
     timeout-minutes: 10
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
+        redis-version: [4, 5, 6]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


ExampleExtractor/config.yml
```diff
@@ -12,7 +12,7 @@ cognite:
     scopes:
       - ${BF_TEST_SCOPE}
 high-availability:
-  index: 1
-  raw:
-    database-name: ${BF_TEST_DB}
+  index: 0
+  redis:
+    connection-string: "localhost"
     table-name: ${BF_TEST_TABLE}
```
einarmo, line 18: I'd prefer if you keep it with raw for now, since that's more example-y.

einarmo (approved):
Neat! I think you'll have to figure out a way to run Redis in Jenkins. Might be an interesting project for your last few days.

.github/workflows/build-and-test.yml
```diff
@@ -10,6 +10,10 @@ jobs:
   build:
     timeout-minutes: 10
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
+        redis-version: [4, 5, 6]
```
einarmo, line 16: You don't use node, and this is overkill. Just remove this whole bit and put in a constant in place of `${{ matrix.redis-version }}`

kjerand: Yeah, I was just testing the example code

PR #190, opened 2022-09-16
Improve HA startup time

einarmo:
It takes a bit long now, there are two reasons:
 - Instead of reusing uploaded state, we re-download the state and use that, which may be outdated.
 - We wait on startup, unnecessarily.

### DIFF
--- a/ExtractorUtils.Test/unit/RawHighAvailabilityTest.cs
+++ b/ExtractorUtils.Test/unit/RawHighAvailabilityTest.cs
@@ -131,16 +131,17 @@ namespace ExtractorUtils.Test.Unit
 
                 Assert.True(extractorManager._state.CurrentState.Count == 0);
 
+                await extractorManager.UploadLogToState();
                 await extractorManager.UpdateExtractorState();
 
                 // Testing that the state has changed.
-                Assert.True(extractorManager._state.CurrentState.Count == 4);
+                Assert.Equal(4, extractorManager._state.CurrentState.Count);
 
                 // Checking that each value in the state is the same as in the db.
                 foreach (RawExtractorInstance instance in extractorManager._state.CurrentState.Values)
                 {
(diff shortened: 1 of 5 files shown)


ExtractorUtils.Test/unit/RawHighAvailabilityTest.cs
```diff
@@ -130,16 +131,17 @@ public async Task TestUpdateExtractorState()
 
                 Assert.True(extractorManager._state.CurrentState.Count == 0);
 
+                await extractorManager.UploadLogToState();
                 await extractorManager.UpdateExtractorState();
 
                 // Testing that the state has changed.
-                Assert.True(extractorManager._state.CurrentState.Count == 4);
+                Assert.Equal(4, extractorManager._state.CurrentState.Count);
```
timurgen, line 138: what is this magic number 4?

ExtractorUtils/HighAvailability/RawHighAvailabilityManager.cs
```diff
@@ -65,26 +69,14 @@ internal override async Task UpdateExtractorState()
                 var rows = await _destination.CogniteClient.Raw.ListRowsAsync<RawLogData>(_config.Raw?.DatabaseName, _config.Raw?.TableName).ConfigureAwait(false);
```
timurgen, line 69: Does this table store all states for all possible extractors?
If so is there a chance we will get too much data we don't need about all "other" extractors? Which seems to be not so good if there is a lot of data

einarmo: It's for High Availability, so an extractor needs to know about all instances of that same extractor. With correct configuration there should only be one small Raw row per standby extractor in that table, so a very small amount of data.

timurgen (approved):
lgtm!

PR #192, opened 2022-10-10
Add alias for pipeline-id, to align with python utils

einarmo:
It is now possible to create a single common base config file for remote python and .NET extractors.


timurgen (approved):
lgtm!

PR #193, opened 2022-10-19
[DOG-367] Fix issue overriding fields in subtype

einarmo:
This occurs if the Cognite field is a subtype of CogniteConfig (which we do support), and no cognite field is created in the remote config.


timurgen (approved):
lgtm!

PR #195, opened 2022-10-28
feat: automatic authenticator implied

toondaey:
This ensures that any config with `tenant` set will always use the `MSALAuthenticator` and that with `token-url` will use the basic `Authenticator` implementation. This renders `idp-authentication.implementation` deprecated.

### DIFF
--- a/ExtractorUtils.Test/integration/AuthIntegrationTest.cs
+++ b/ExtractorUtils.Test/integration/AuthIntegrationTest.cs
@@ -50,7 +50,6 @@ namespace ExtractorUtils.Test.Integration
 
                 var configList = new List<string>(CDFTester.GetConfig(host));
                 configList.Insert(13, "    implementation: basic");
-                configList.Insert(14, "    token-url: https://login.microsoftonline.com/${BF_TEST_TENANT}/oauth2/v2.0/token");
                 var configBasic = configList.ToArray();
                 using (var tester = new CDFTester(configBasic, _output))
                 {
--- a/ExtractorUtils/Cognite/DestinationUtils.cs
+++ b/ExtractorUtils/Cognite/DestinationUtils.cs
@@ -250,12 +250,6 @@ namespace Cognite.Extractor.Utils
                     "Cannot configure Builder: Only either of 'idp-authentication.tenant' or 'idp-authentication.token-url' can be set"
                 );
             }
-            else if (String.IsNullOrWhiteSpace(_tenant) && String.IsNullOrWhiteSpace(_tokenUrl))
-            {
-                throw new CogniteUtilsException(
-                    "Cannot configure Builder: Either one of 'idp-authentication.tenant' or 'idp-authentication.token-url' has to be set"
-                );
-            }
             else if (!String.IsNullOrWhiteSpace(_tenant) && String.IsNullOrWhiteSpace(config.IdpAuthentication?.Authority))
             {
                 throw new CogniteUtilsException(
--- a/ExtractorUtils/Cognite/DestinationUtils.cs
+++ b/ExtractorUtils/Cognite/DestinationUtils.cs
@@ -247,7 +247,7 @@ namespace Cognite.Extractor.Utils
             if (!(tenant is null) && !(tokenUrl is null))
             {
                 throw new CogniteUtilsException(
-                    "Cannot configure Builder: Only either 'idp-authentication.tenant' or 'idp-authentication.token-url' can be set"
+                    "Cannot configure Builder: Only one of 'idp-authentication.tenant' or 'idp-authentication.token-url' can be set"
                 );
             }
             else if (!(tenant is null) && config.IdpAuthentication?.Authority is null)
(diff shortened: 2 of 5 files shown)


ExtractorUtils/Cognite/DestinationUtils.cs
```diff
@@ -242,6 +240,23 @@ public static void ConfigureSslPolicy(CertificateConfig config)
             {
                 throw new CogniteUtilsException("Cannot configure Builder: Project is not configured");
             }
+
+            string? _tenant = config.IdpAuthentication?.Tenant.TrimToNull();
+            string? _tokenUrl = config.IdpAuthentication?.TokenUrl.TrimToNull();
+
+            if (!String.IsNullOrWhiteSpace(_tenant) && !String.IsNullOrWhiteSpace(_tokenUrl))
```
einarmo, line 247: Use `string` over `String`, please.

ExtractorUtils/Cognite/DestinationUtils.cs
```diff
@@ -242,6 +240,23 @@ public static void ConfigureSslPolicy(CertificateConfig config)
...
+            {
+                throw new CogniteUtilsException(
+                    "Cannot configure Builder: Only either 'idp-authentication.tenant' or 'idp-authentication.token-url' can be set"
```
einarmo, line 250: "Only one of", perhaps?

ExtractorUtils/config/config.example.yml
```diff
@@ -36,9 +36,10 @@ cognite:
         # MSAL (recommended) - Microsoft Authentication Library, Works only with authority/tenant
         # Basic - Post to authentication endpoint and parse JSON response, works with both authority/tenant and token-url
         # Default is MSAL
-        audience:
-        # Audience
+        # Deprecated: implementation is implied based on either token-url or tenant
```
einarmo, line 39: Remove it entirely, we don't want to include deprecated options in here.

einarmo: Especially since it no longer works.

ExtractorUtils.Test/integration/AuthIntegrationTest.cs
```diff
@@ -50,7 +50,6 @@ public async Task TestClientConfig(CogniteHost host)
 
                 var configList = new List<string>(CDFTester.GetConfig(host));
                 configList.Insert(13, "    implementation: basic");
-                configList.Insert(14, "    token-url: https://login.microsoftonline.com/${BF_TEST_TENANT}/oauth2/v2.0/token");
```
einarmo, line 53: Why is this gone, what does this part of the test do now? Since `implementation: basic` does nothing.

ExtractorUtils/Cognite/DestinationUtils.cs
einarmo: Also, since you already used `TrimToNull` you can just compare it with `null` here: `_tenant is null`.

ExtractorUtils/Cognite/DestinationUtils.cs
```diff
@@ -242,6 +240,23 @@ public static void ConfigureSslPolicy(CertificateConfig config)
             {
                 throw new CogniteUtilsException("Cannot configure Builder: Project is not configured");
             }
+
+            string? _tenant = config.IdpAuthentication?.Tenant.TrimToNull();
+            string? _tokenUrl = config.IdpAuthentication?.TokenUrl.TrimToNull();
```
einarmo, line 245: underscore prefixed variables are reserved for private members, use `tenant` and `tokenUrl` instead.

ExtractorUtils/Cognite/DestinationUtils.cs
```diff
@@ -242,6 +240,23 @@ public static void ConfigureSslPolicy(CertificateConfig config)
...
+            if (!String.IsNullOrWhiteSpace(_tenant) && !String.IsNullOrWhiteSpace(_tokenUrl))
+            {
+                throw new CogniteUtilsException(
+                    "Cannot configure Builder: Only either 'idp-authentication.tenant' or 'idp-authentication.token-url' can be set"
+                );
+            }
+            else if (!String.IsNullOrWhiteSpace(_tenant) && String.IsNullOrWhiteSpace(config.IdpAuthentication?.Authority))
+            {
+                throw new CogniteUtilsException(
+                    "Cannot configure Builder: The 'idp-authentication.authority' is required when 'idp-authentication.tenant' is provided"
```
einarmo, line 256: Remove "The" here.

einarmo (changes requested):
Some minor issues.

ExtractorUtils.Test/integration/AuthIntegrationTest.cs
toondaey: Since the config this appends to the list of config [here](https://github.com/cognitedata/dotnet-extractor-utils/blob/9f15731a666570967923ccee8871f9f22b4c8023/ExtractorUtils.Test/CDFTester.cs#L102) and there's a check that doesn't allow setting both tenant and token-url, this always throw an error hence the removal.

einarmo (approved):
LGTM

PR #196, opened 2022-11-01
Schema

einarmo:
Author schema definitions for extractor utils.

We should look into a way to host these publicly, so that we can retrieve them as part of CICD.

To make writing these bearable, we must be able to split them into separate files. When storing them in the database we probably would like to download and merge them. A solution could be to use the `$defs` keyword:

 - Scan the document for `$ref`s
 - Load all referenced documents, either from the internet or from local files.
 - Store each referenced item in `$defs`.
 - Replace remote refs with local refs in generated document.

This is doable, and relatively clean.

The schema can be referenced quite elegantly from github, so long as we tag releases:

`https://github.com/cognitedata/dotnet-extractor-utils/blob/v1.4.0/README.md?raw=true` is a public, plaintext reference to the readme file at version 1.4.0.

### DIFF
--- a/ExampleExtractor/config.yml
+++ b/ExampleExtractor/config.yml
@@ -13,9 +13,12 @@ cognite:
     scopes:
       - ${BF_TEST_SCOPE}
   extraction-pipeline:
-    pipeline-id: net-remote-test
+    external-id: net-remote-test
 high-availability:
   index: 1
   raw:
     database-name: ${BF_TEST_DB}
     table-name: ${BF_TEST_TABLE}
+logger:
+  console:
+    level: debug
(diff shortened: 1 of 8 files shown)


toondaey (approved):
🚀

toondaey:
I think we can create a repository specifically for the purpose of hosting the (generated) schema publicly.

einarmo:
> I think we can create a repository specifically for the purpose of hosting the (generated) schema publicly.

I don't really think we need to? We don't need to merge this schema, so we can just keep them in this repo.

Look at https://github.com/cognitedata/extractor-uploader/pull/13 for a system for bundling that manages to handle references to `base_config.schema.json`

PR #197, opened 2022-11-03
Remove top level unevaluatedProperties

einarmo:
They make it so that we're unable to extend them (which makes sense). Any types which we intend users to extend we drop unevaluatedProperties from. This still works because we just let users add those directly on their own properties if they use them directly.

See ExampleExtractor/full_config.schema.json for an example, we put unevaluatedProperties on the top level, and inside each externally referenced property, indicating that we won't extend them further.

In theory you could reference properties from other types directly, and extending those would work poorly, but that's a sacrifice we can make.


timurgen (approved):
lgtm!

PR #198, opened 2022-11-04
Add support for parsing null values from YAML

ozangoktan:
This PR updates YAML deserializer to support null value parsing. Possible null representations are meant to cover the options from https://yaml.org/type/null.html.

### DIFF
--- a/Cognite.Config/Configuration.cs
+++ b/Cognite.Config/Configuration.cs
@@ -319,13 +319,13 @@ namespace Cognite.Extractor.Configuration
         {
         }
 
-        private static HashSet<string> _nullLiteralValues { get; set; } = new HashSet<string>() { "null", "NULL", "Null", "", "~" };
-
         private static bool IsNumericType(Type t)
         {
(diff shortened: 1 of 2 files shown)


einarmo:
This issue confused me a little. We really should handle nulls. YamlDotNet custom serialization does not have that much documentation, so I went digging a bit, and found that the issue is that we _consume_ all string fields in the custom serializer, whether or not they are a match for the regex.

What we want to do is only consume fields that do match the regex. I.e. 

```
if (parser.Accept<Scalar>(out var scalar) && scalar != null && _envRegex.IsMatch(scalar.Value))
{
    parser.MoveNext();
    value = _envRegex.Replace(scalar.Value, LookupEnvironment);
    return true;
}
```

I've pushed a commit to that effect. Please verify that it works for your issue as well.

Thanks for pointing this out and writing a test, it might have had other consequences as well.

timurgen (approved):
lgtm!

ozangoktan:
I can confirm the updated method works for my use case. No problem!

PR #201, opened 2022-11-14
Handle infinite timespan

einarmo:
Normal comparison doesn't handle this, because it's an extension from timeout. We generally care about it.


toondaey (approved):
lgtm

PR #202, opened 2022-11-21
Fix retry counting

einarmo:
The number of maximum retries should be equal to max-retries, and the logged attempt number should start at 1.


toondaey (approved):
lgtm

PR #204, opened 2023-01-18
Reload remote config after restart

einarmo:
The extractor would not properly enable remote config after restarting, since it kept the previous config. This meant that config changes were not picked up. This should fix that, making it so that if we enable remote config once, it is enabled in all future iterations as well.

### DIFF
--- a/ExtractorUtils/ExtractorRunner.cs
+++ b/ExtractorUtils/ExtractorRunner.cs
@@ -232,6 +232,8 @@ namespace Cognite.Extractor.Utils
                 catch { }
             }
 
+            bool usesRemoteConfig = false;
+
             Console.CancelKeyPress += CancelKeyPressHandler;
             while (!source.IsCancellationRequested)
             {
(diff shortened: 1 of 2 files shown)


ExtractorUtils/ExtractorRunner.cs
```diff
@@ -245,8 +247,7 @@ void CancelKeyPressHandler(object sender, ConsoleCancelEventArgs eArgs)
                 ConfigurationException? exception = null;
                 try
                 {
-                    bool usesRemoteConfig = false;
-                    if (options.AllowRemoteConfig && options.Config == null)
+                    if (options.AllowRemoteConfig && options.Config == null || usesRemoteConfig)
```
toondaey, line 250: I think the `usesRemoteConfig` here will always evaluate to `false` so checking it in the conditional might be redundant.

einarmo: It won't any more, I moved the bool outside the loop.

First iteration of the loop it is set to `true`.

toondaey: I didn't notice the loop, my bad.

toondaey (approved):

PR #238, opened 2023-03-22
Certificate authentication

einarmo:
This required adding .net 7 as a target, which caused a ton of warnings. This also fixes those.


toondaey (approved):
lgtm

PR #257, opened 2023-05-25
Remove api keys

einarmo:


toondaey (commented):
We should probably update the readme as well, removing all the references to API keys.
Also found a reference [here](https://github.com/cognitedata/dotnet-extractor-utils/blob/f1e660dec226b9ca7a968e6a3ea98794cc5564ab/ExtractorUtils/Cognite/DestinationUtils.cs#L219) and [here](https://github.com/cognitedata/dotnet-extractor-utils/blob/f1e660dec226b9ca7a968e6a3ea98794cc5564ab/docfx_project/tutorials/intro.md#L39)

toondaey (approved):
👍🏾

PR #272, opened 2023-06-28
Replace Cronos with NCrontab

einarmo:
It isn't any better, but there exists a strongly named version, so we don't have to strongly name it ourself


mathialo (approved):

toondaey (approved):
👍🏾

PR #282, opened 2023-07-25
Upload queue improvements

einarmo:
- Better handle fatal errors when retrying.
 - Avoid mixing sync and async in BaseUploadQueue
 - Return the list of failed items to the user, when possible.
 - Make the BaseUploadQueue constructor public

### DIFF
--- a/Cognite.Extensions/CogniteResult.cs
+++ b/Cognite.Extensions/CogniteResult.cs
@@ -436,6 +436,11 @@ namespace Cognite.Extensions
         {
             return new CogniteResult<TResult, TRep>(Errors?.Select(e => e.ReplaceSkipped(replace)), Results);
         }
+
+        /// <summary>
+        /// All items that were skipped in this result. Found by joining all errors.
+        /// </summary>
+        public IEnumerable<TError> AllSkipped => Errors?.Where(e => e.Skipped != null)?.SelectMany(e => e.Skipped!) ?? Enumerable.Empty<TError>();
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


Cognite.Extensions/CogniteResult.cs
```diff
@@ -436,6 +436,11 @@ public CogniteResult(IEnumerable<CogniteError<TError>>? errors, IEnumerable<TRes
         {
             return new CogniteResult<TResult, TRep>(Errors?.Select(e => e.ReplaceSkipped(replace)), Results);
         }
+
+        /// <summary>
+        /// All items that were skipped in this result. Found by joining all errors.
```
ozangoktan, line 441: `joining all errors` seems a bit incorrect

Cognite.Extensions/TimeSeries/DataPointExtensions.cs
```diff
@@ -302,14 +302,18 @@ public static DataPointInsertionRequest ToInsertRequest(this IDictionary<Identit
...
                     if (error.Type == ErrorType.FatalFailure
                         && (retryMode == RetryMode.OnFatal
                             || retryMode == RetryMode.OnFatalKeepDuplicates))
                     {
                         await Task.Delay(1000, token).ConfigureAwait(false);
                     }
-                    else if (retryMode == RetryMode.None) break;
+                    else if (retryMode == RetryMode.None)
+                    {
+                        errors.Add(error);
```
ozangoktan, line 314: error.Complete check is removed here, is this intended? (Based on similarity of changes to this file and Cognite.Extensions/Sequences/SequenceExtensions.cs)

einarmo: It is intentional, yeah, if it's not removed in SequenceExtensions then the problem is there.

ozangoktan (approved):
LGTM

PR #294, opened 2023-08-30
Thread-proof authenticators

einarmo:
They are actually a little racy, this should help make them less prone to duplicate token requests.


ozangoktan (approved):
LGTM

PR #299, opened 2023-09-06
Add utility method for performing atomic updates

einarmo:
Needed for some stuff in the OPC-UA extractor.

### DIFF
--- a/Cognite.Extensions/DataModels/AtomicUpdateUtils.cs
+++ b/Cognite.Extensions/DataModels/AtomicUpdateUtils.cs
@@ -28,6 +28,8 @@ namespace Cognite.Extensions.DataModels
         /// should create a write by updating the given instances.
         /// May be called multiple times if the upsert fails with a conflict.</param>
         /// <param name="token">Cancellation token</param>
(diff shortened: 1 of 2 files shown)


Cognite.Extensions/DataModels/AtomicUpdateUtils.cs
```diff
@@ -0,0 +1,97 @@
...
+                        AutoCreateStartNodes = true,
+                    }, token).ConfigureAwait(false);
+                    return res;
+                }
+                catch (ResponseException ex)
+                {
+                    if (ex.Code == 409 && ex.Message == "A version conflict caused the ingest to fail.")
+                    {
+                        // Try again right away
+                        continue;
```
ozangoktan, line 89: Not sure how many times this can happen, but there should be an upper limit for retries.

einarmo: Yeah, sure, it should be unlikely to happen at all, but a cap is probably smart...

ozangoktan (approved):
LGTM

PR #320, opened 2023-11-07
Better remote configuration

einarmo:
Create a config file that can be used for remote config. This also makes a few changes to make that more possible.

 - Allow scopes to be configured either as a list of strings, or as space separated strings.
 - Allow an empty "certificate" block
 - Improve handling of empty values.

### DIFF
--- a/ExtractorUtils/config/remote_config.yml
+++ b/ExtractorUtils/config/remote_config.yml
@@ -1,0 +1,39 @@
+# Sample configuration file for remote config. This can be used as remote config directly.
+
+cognite:
+    # The project to connect to in the API
+    project: ${COGNITE_PROJECT}
+    # Cognite service url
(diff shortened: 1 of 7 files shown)


ExtractorUtils/config/remote_config.yml
```diff
@@ -0,0 +1,39 @@
+# Sample configuration file for remote config. This can be used as remote config directly.
```
toondaey, line 1: Could this file have been included by mistake? Doesn't seem to be used.

ozangoktan (approved):
LGTM

ExtractorUtils/config/remote_config.yml
einarmo: It's a reference file, meant to be what this PR makes possible.

PR #332, opened 2023-11-16
Add option to set blocking resource counter capacity

einarmo:


Cognite.Common/BlockingResourceCounter.cs
```diff
@@ -72,13 +78,34 @@ public Task<int> Take(int count, bool shouldBlock)
...
+                Count += diff;
+
+                Monitor.PulseAll(_lock);
+            }
+        }
+
         /// <inheritdocs />
         public void Free(int count)
         {
+            if (count < 0) throw new ArgumentException("count must be greater than or equal to 0");
```
ozangoktan, line 104:
```suggestion
            if (count < 0) throw new ArgumentException("count must be greater than or equal to 0");
            if (count == 0) return;
```

ozangoktan (approved):
LGTM, added a suggestion that might be a good addition

PR #352, opened 2024-01-02
Try addressing test flakiness

ozangoktan:


einarmo (approved):
Slightly hard to tell what actually changed, but this is just using the xunit test output instead of the console, right? It might help?

ozangoktan:
> Slightly hard to tell what actually changed, but this is just using the xunit test output instead of the console, right? It might help?

Pretty much, yes.

PR #362, opened 2024-01-16
KeyVault support

einarmo:
Designed to be as similar as possible to the python version, which required some work.

Among other things, it required a custom enum serializer to let us have aliases. This is a neat addition anyway.

This also makes the snk path relative, since solutiondir often doesn't exist.

### DIFF
--- a/Cognite.Config/KeyVault.cs
+++ b/Cognite.Config/KeyVault.cs
@@ -1,0 +1,142 @@
+using System;
+using System.Runtime.Serialization;
+using Azure.Core;
+using Azure.Identity;
+using Azure.Security.KeyVault.Secrets;
+using Cognite.Extractor.Common;
+using Cognite.Extractor.Configuration;
+using YamlDotNet.Core;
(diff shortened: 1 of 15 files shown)


Cognite.Config/KeyVault.cs
```diff
@@ -0,0 +1,142 @@
...
+    public class KeyVaultConfig
+    {
+        /// <summary>
+        /// How to authenticate against key vault.
+        /// </summary>
+        public KeyVaultAuthenticationMethod? AuthenticationMethod { get; set; }
+        /// <summary>
+        /// Azure key vault name.
+        /// </summary>
+        public string? VaultName { get; set; }
```
mathialo, line 44: Nitpick: This property is called [`keyvault-name` in the python utils](https://github.com/cognitedata/python-extractor-utils/blob/master/cognite/extractorutils/configtools/loaders.py#L73), we should use the same name here

mathialo (approved):

PR #405, opened 2024-04-17
Support for bad/missing values and status codes on strings

einarmo:


Cognite.Extensions/TimeSeries/DataPointSanitation.cs
```diff
@@ -18,25 +18,29 @@ public static Datapoint Sanitize(this Datapoint point, double? nanReplacement)
         {
             if (point.IsString)
             {
-                if (point.StringValue == null || point.StringValue.Length > CogniteUtils.StringLengthMax)
+                if (point.StringValue == null && point.Status.IsGood || (point.StringValue?.Length ?? 0) > CogniteUtils.StringLengthMax)
                 {
-                    return new Datapoint(point.Timestamp, point.StringValue.Truncate(CogniteUtils.StringLengthMax) ?? "");
+                    return new Datapoint(point.Timestamp, point.StringValue.Truncate(CogniteUtils.StringLengthMax) ?? "", point.Status);
```
ozangoktan, line 23: Nit: Would prefer if the logic was split apart along the OR

ozangoktan (approved):
LGTM, added a minor comment which can improve readability a bit

PR #411, opened 2024-05-07
Fix issue with token renewal

einarmo:
The token would be stored as renewed even if the request itself failed, which is very wrong.


rsjr (approved):
aprobasjon

PR #461, opened 2024-08-30
Add Support for CDM TimeSeries

ozangoktan:
- Add support for writing to CDM TimeSeries as an alternative destination. 
- Remove status codes support that is already moved to the dotnet SDK


Cognite.Extensions/CogniteResult.cs
```diff
@@ -752,7 +799,15 @@ public enum ResourceType
...
-        None = -1
+        None = -1,
+        /// <summary>
+        /// Data modeling space id
+        /// </summary>
+        SpaceId,
+        /// <summary>
+        /// Data modeling instance id
+        /// </summary>
+        InstanceId
```
einarmo, line 810: Move these above None, just prefer to have that at the bottom still.

einarmo (commented):
The docs should mention that the IDM stuff is in beta.

Mostly good though, tests build my confidence.

Cognite.Extensions/CogniteResult.cs
ozangoktan: The reason I put them below None is because C# enums are assigned numerical values based on their order when not explicitly assigned a numerical value, so it is breaking to change the order.

einarmo: Yes, but `None` _has_ a numerical value.

ozangoktan: Ah, true, I will pop those up

einarmo (approved):
Big PR, hard to say if everything is exactly correct, but it looks fine.

We may want to find a way to make the whole extensions bit more generic, though I expect it to be a major pain.

PR #483, opened 2024-10-23
Enable code analyzers in ExtractorUtils.Test

ozangoktan:


einarmo (commented):
Why do we actually need this?

ozangoktan:
> Why do we actually need this?

The original form was indeed not necessary and probably would not have worked either. I've salvaged the parts that can be considered an improvement.

einarmo (approved):
Sure, this is an improvement I think.

PR #487, opened 2024-10-25
Improve sanitation error log

einarmo:


Cognite.Extensions/LoggerExtensions.cs
```diff
@@ -46,11 +46,15 @@ public static void LogCogniteError<TError>(this ILogger logger,
                     cogniteString += $", Message: {error.Exception.Message}";
                 }
             }
+            else if (error.Message != null)
+            {
+                cogniteString = $" {error.Message}";
```
ozangoktan, line 51: Are Exception and Message fields mutually exclusive? We should probably include both if they are available.

ozangoktan (approved):

PR #488, opened 2024-10-31
Add support for CDM Assets

ozangoktan:
Looking for input on what to do with the specialized counters and enum types per CDM type. Currently I created a  common one for DM instances for a couple of reasons:

- Creating a new counter for each new CDM type seems counter-intuitive and is likely to get cumbersome as we add more. Unifying them would asset-centric existing ones would work, but would need some changes. 
- Injecting the relevant counters and result types into generic methods would mean moving the counters into CDM resources, which live on the SDK or adding yet another layer of abstraction here. 

I will be sending in a commit that does the the additional layer of abstraction, but sending in a draft PR now as to get more input on this.


einarmo:
Could we just add a label to the metric?

Cognite.Extensions/CdfMetrics.cs
```diff
@@ -8,8 +9,10 @@ static class CdfMetrics
             "Number and duration of asset requests to CDF", "endpoint");
         public static Summary TimeSeries { get; } = Metrics.CreateSummary("extractor_utils_cdf_timeseries_requests",
             "Number and duration of time-series requests to CDF", "endpoint");
-        public static Summary CoreTimeSeries { get; } = Metrics.CreateSummary("extractor_utils_cdf_core_timeseries_requests",
-            "Number and duration of core data model time-series requests to CDF", "endpoint");
+        public static Summary.Child Instances(ViewIdentifier view, string operation)
+        {
+            return Metrics.CreateSummary("extractor_utils_cdf_instances_requests", "Number and duration of instance requests CDF data modeling", "endpoint", "view_identifier").WithLabels(operation, $"{view.Space}-{view.ExternalId}");
+        }
```
einarmo, line 15: This feels very wrong. Now you are creating a new metrics object every time you call this endpoint

ozangoktan: 5509a7b6ad94e8dcce9d8b9d3bda5137a9e87dd6 should address that

einarmo (commented):
I wrote the comment but forgot to submit it. Would be good to fix the plurality of the new asset methods.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/ExtractorAssetExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <param name="asset">Cognite asset resource</param>
+        /// <param name="instanceIds">Instance Ids</param>
+        /// <param name="buildAsset">Function that builds CogniteSdk SourcedNodeWrite objects</param>
+        /// <param name="chunkSize">Chunk size</param>
+        /// <param name="throttleSize">Throttle size</param>
+        /// <param name="retryMode">How to handle failed requests</param>
+        /// <param name="sanitationMode">The type of sanitation to apply to asset before creating</param>
+        /// <param name="token">Cancellation token</param>
+        /// <returns>A <see cref="CogniteResult{TResult, TError}"/> containing errors that occured and a list of the created and found asset</returns>
+        public static Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateAssetAsync<T>(
```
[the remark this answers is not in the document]
ozangoktan: Did some cleanup with 094501deacf5dc3095a967ddb66d2d0fe74dc425, should address this as well

einarmo (approved):
Complex PR, but I think this is good, looks like a nice improvement.

PR #506, opened 2025-01-09
Start work on the unstable module

einarmo:
Defines `ConnectionConfig` and some utils related to this. Lots more to do before we can actually use this for anything.


ozangoktan (approved):
LGTM

PR #513, opened 2025-01-28
Initial impl of task scheduler for integrations

einarmo:
Class that schedules tasks in the way we expect for integrations, reporting errors, warnings, start, and end.

### DIFF
--- a/ExtractorUtils/Unstable/Tasks/Task.cs
+++ b/ExtractorUtils/Unstable/Tasks/Task.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using CogniteSdk.Alpha;
 
 namespace Cognite.ExtractorUtils.Unstable.Tasks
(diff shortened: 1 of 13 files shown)


ExtractorUtils/Unstable/Tasks/Task.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /// <summary>
+        /// Report that this task has started.
+        /// </summary>
+        /// <param name="timestamp"></param>
+        public void ReportStart(DateTime? timestamp = null)
+        {
+            if (_isRunning) throw new InvalidOperationException("Attempted to start task that is already running");
+            _isRunning = true;
```
ozangoktan, line 42: Probably should wrap these 2 lines and the counterpart in ReportEnd with a lock, or perhaps the entire method contents.

einarmo: I really don't think it's necessary. Hitting this is a bug in the extractor or elsewhere, and it will throw a slightly less clear error further down the line. The extremely unlikely scenario of a race condition here IMO isn't worth the lock.

ozangoktan (approved):
LGTM

PR #523, opened 2025-03-06
Improve the timestamp parsing

einarmo:
Add a timestampwrapper config utility, which should make this type easier to use in a good way in config.

Make it accept RFC3339-like datetimes, but also shortened. We require a timezone, which IMO is a good idea, to avoid ambiguity.


ozangoktan (dismissed):
Looks great!

ozangoktan (approved):

ExtractorUtils.Test/unit/CogniteTimeTest.cs
```diff
@@ -242,6 +245,31 @@ public static void TestParseTimeAbsolute()
...
         }
+
+        [Theory]
+        [InlineData("2020-01-01T01:00:00+01:00", false)]
+        [InlineData("2020-01-01T00:00:00Z", false)]
+        [InlineData("2020-01-01T00:00:00+00:00", false)]
+        [InlineData("2020-01-01T01:00+01:00", false)]
+        [InlineData("2020-01-01T01+01:00", false)]
+        [InlineData("2020-01-01+00:00", false)]
+        [InlineData("2020-01-01Z", false)]
```
finnag, line 256: That was an extremely simplified list of tests?

In a good old test driven methology the winning implementation would be `return DateTime(2020, 1, 1, 0, 0, 0, DateTimeKind.Utc)` to satisfy all tests.

We should at least test that it parses the numbers in the right order (day/month/year/hour/minute/), that time zones work as intended, that bad dates do NOT parse, and so on.

finnag (changes requested):
This deserves a little more thorough testing

ExtractorUtils.Test/unit/CogniteTimeTest.cs
einarmo: I added some more test cases, but there really is only so much value in testing a standard library function.

ozangoktan (approved):

ExtractorUtils.Test/unit/CogniteTimeTest.cs
einarmo: @finnag I've updated this, it's ready for another risk review.

finnag (approved):
:unicorn:7

PR #527, opened 2025-03-18
Add a config wrapper, fix namespaces

einarmo:
- Add a config wrapper, letting us pass the revision number to the extractor. This is necessary for later when the extractor will restart based on responses from checkin.
 - Fix incorrect namespaces. The base namespace is Cognite.Extractor.Utils, not Cognite.ExtractorUtils


ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -132,6 +161,34 @@ private bool ShouldLoadNewConfig(int? targetRevision, ConfigMode mode)
...
+        public ConfigWrapper<T> GetConfigWrapper()
+        {
+            int? revision;
+            if (_state.Mode == ConfigMode.Local)
+            {
+                revision = null;
+            }
+            else if (_state.CurrentRevision == null)
+            {
+                throw new InvalidOperationException("Attempt to resolve config when no revision is set.");
```
ozangoktan, line 183:
```suggestion
                throw new InvalidOperationException("Attempt to resolve remote config without revision.");
```

ozangoktan (approved):

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -132,6 +161,34 @@ private bool ShouldLoadNewConfig(int? targetRevision, ConfigMode mode)
...
+            }
+            else if (_state.CurrentRevision == null)
+            {
+                throw new InvalidOperationException("Attempt to resolve remote config without revision.");
+            }
+            else
+            {
+                revision = _state.CurrentRevision;
+            }
+            return new ConfigWrapper<T>(_state.Config!, revision);
```
rbtcollins, line 189: Assuming you keep the types... an early return here makes it clearer that there are three exits. 

```suggestion
            int? revision;
            if (_state.Mode == ConfigMode.Local)
            {
                new ConfigWrapper<T>(_state.Config!,  null);
            }
            else if (_state.CurrentRevision == null)
            {
                throw new InvalidOperationException("Attempt to resolve remote config without revision.");
            }
            else
            {
                new ConfigWrapper<T>(_state.Config!, _state.CurrentRevision);
            }
```

Better (IMO) representing the logic chain as a table:

```suggestion
       int? revision = (_state.Mode, _state.CurrentRevision) switch {
            (=ConfigMode.Local, _) => null,
            (_, is null) =>throw new InvalidOperationException("Attempt to resolve remote config without revision.");
            (_, revision) => revision
        }
        return new ConfigWrapper<T>(_state.Config!, revision);
```

The logic is unchanged in these cases, and I don't think its all that hard to tell with the consolidated exit, but I do think either of these versions is easier to read.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -132,6 +161,34 @@ private bool ShouldLoadNewConfig(int? targetRevision, ConfigMode mode)
...
+        /// 
+        /// The config wrapper contains the current revision or "null" if the config
+        /// is local.
+        /// </summary>
+        /// <returns></returns>
+        /// <exception cref="InvalidOperationException">If no config is set.</exception>
+        public ConfigWrapper<T> GetConfigWrapper()
+        {
+            int? revision;
+            if (_state.Mode == ConfigMode.Local)
```
rbtcollins, line 177: I smell something a little off about the types here. You have a CurrentRevision which is not applical for ConfigMode.Local. That then depends on using it the right way, because it can represent invalid states: ConfigMode.Local= true with CurrentRevision is not null

And the code in this method is then setting some boundaries around this.

But this is going to spread through your code: everything that wants to behave differently based on local / remote configuration

I'm just sketching here, but in Rust I would look for

```
/// The configuration of the extract.
enum Config<T> {
    Local(T),
    Remote{int revision, config: T}
}
```

I'm no csharp expert, but https://microsoft.github.io/rust-for-dotnet-devs/latest/language/custom-types/enums.html which I read a while back suggests that an analog would be:

```
abstract record Config<T>
{
    public sealed record Local<T>(T config): Config<T>;
    public sealed record Remote<T>(int revision, T config): Config<T>;
}
```

but ... we don't get missing case detection in switch statement :(.

so perhaps thats a bad idea. What I think is important though, is to think about how you can manage these two things:
1) make sure code won't be operating on invalid states.
2) see if you can limit the ability to represent invalid states

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -132,6 +161,34 @@ private bool ShouldLoadNewConfig(int? targetRevision, ConfigMode mode)
...
+            {
+                revision = null;
+            }
+            else if (_state.CurrentRevision == null)
+            {
+                throw new InvalidOperationException("Attempt to resolve remote config without revision.");
+            }
+            else
+            {
+                revision = _state.CurrentRevision;
```
rbtcollins, line 187: This if sequence is missing a case for ConfigMode.None. See the earlier comment about invalid states. I realise ConfigMode is an existing type, but ConfigMode.None seems like an attractive nuisance : it leaks the late initialisation of the config through the whole codebase.

I'm assuming you're heading towards an online-reconfiguration pattern where the config will get a new revision, and then everything in the extract that depends on the current configuration will get a signal, and reload etc.

This might be easier to model if instead of revision = null meaning 'local configuration', revision = null means 'program is initialising', and for local configuration you synthesis a revision int from the number of times the config file is observed to change.

Then you have null = unconfigured, 0 = local file read and parsed / first remote config, 1 = local file was updated, or new remote config, etc...)

If you need to dynamically switch between local and remote configuration, a similar pattern could deliver that, but I don't want to proposed designs for a thing you're not planning on.

rbtcollins (commented):
I think there is a modelling / type hazard here, happy to discuss.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
einarmo: There are some limitations to what we can do since this needs to support .NET framework, which hasn't had new features added in a decade, so just keep that in mind when it comes to fancy new C# features.

It may be that it would be better to replace the entire wrapper type with some sort of type inheritance. I just really try to avoid that as it can make it impossible to figure out what code is _actually_ being executed.

einarmo: Also, I'll try not to complain too much, but huge suggestions like this hurt a _ton_ when they come a week late. I'll have to throw away 2k lines of code if I need to rewrite this now.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
einarmo: >  for local configuration you synthesis a revision int

I'm not doing this. It's asking for trouble. A revision number has specific, semantic meaning for the integrations API. This would mean an extractor could report that it was running a specific revision that exists, but lie to the user, which is terrible.

In the end we will have to have a null revision. Storing the config mode in addition to that is just adding an unnecessary invariant.

It's also not possible. Extractors have no way to have persistent state about this, so there's no way to know how many times the file has been read, seeing as the extractor can restart between reads of the config file.

einarmo:
This kind of review really hurts. The logic you're asking us to rewrite has already been through risk review. If we cannot trust that already reviewed code is stable, how do we work at all?

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
einarmo: > Do we have a pathway to something sustainable in future?

Nope. Unfortunately. We're stuck so long as we support PI (and we absolutely have to support PI).

I've now rewritten the wrapper. I don't know if it's better or not, it may be. The state thing is gone completely, as is ConfigMode.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
rbtcollins: Where is the API doc for the integrations API? I can see one for Simulator integration API - is that the same ?

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
einarmo: In general the rewrite shoves the complexity of dealing with the different sources of config up to the caller, which hasn't been written yet. It remains to be seen whether that will be a good thing or not.

einarmo:
I've now rewritten the config wrapper to be a set of classes inheriting from a shared abstract base class. It may be better, I find it hard to say without having written the code that will actually consume this type.

This means that the ConfigMode and ConfigState types are completely gone, instead the mode is implicit from the type. There is still some leakage of the abstraction (we still need to ask for a revision, that's just the way this is supposed to work in general), but there's less switching of the different states.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -4,142 +4,125 @@
...
-namespace Cognite.ExtractorUtils.Unstable.Configuration
+namespace Cognite.Extractor.Utils.Unstable.Configuration
 {
-    class ConfigState<T> where T : VersionedConfig
+    /// <summary>
+    /// Wrapper around a config file with
+    /// extra information about the active revision.
+    /// </summary>
+    /// <typeparam name="TConfig"></typeparam>
+    public class ConfigWrapper<TConfig>
```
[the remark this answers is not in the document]
einarmo: It's technically `VersionedConfig` where it's constructed. It doesn't _use_ that interface though, so I don't necessarily think it's helpful to add a restriction here? Just makes for more verbose code everywhere this is used.

rbtcollins:
:unicorn:

ExtractorUtils/ExtractorUtils.csproj
```diff
@@ -7,7 +7,7 @@
     <Company>Cognite AS</Company>
     <Copyright>Cognite AS</Copyright>
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
-    <RootNamespace>Cognite.ExtractorUtils</RootNamespace>
+    <RootNamespace>Cognite.Extractor.Utils</RootNamespace>
```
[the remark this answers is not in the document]
einarmo: I can do it as a separate PR.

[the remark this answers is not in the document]
einarmo: Also I'm not sure I agree with `TreatWarningsAsErrors` on a csproj level, I can set it in CI, but it's really annoying to not be allowed to have temporarily undocumented or unreferenced code.

ozangoktan (approved):

ExtractorUtils/ExtractorUtils.csproj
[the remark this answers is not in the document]
dbrattli: You can btw fine tune it using: WarningsAsErrors and WarningsNotAsErrors

PR #528, opened 2025-03-19
Add constructor to make a TimestampWrapper from a datetime

einarmo:
This is really handy for tests, and is a bit of a gap. Since this type stores the datetime as a string, we need to convert it to a string, which should be easy enough to do.


HMEiding (dismissed):
LGTM - currently shadowing @dbrattli.

dbrattli (approved):

einarmo:
Love the approvals, but I need a unicorn. @dbrattli , @HMEiding.

HMEiding (approved):
:unicorn:

PR #529, opened 2025-03-24
Improve the checkin worker a bit, add a callback for config changes

einarmo:
This is necessary step, making the checkin worker capable of calling some callback to inform the runtime and/or extractor (not sure yet) that a new config revision has been produced, and that the extractor should restart.

It will be up to the outer loop to decide whether the config changes are significant enough to warrant a change (if we decide to go that way).

I also improved the checkin worker a bit in general, making it a bit more resilient to cancellation.

It's quite important that errors aren't lost when the extractor shuts down, so we need to make sure that any errors that failed to write or weren't written are kept for a future checkin.

### DIFF
--- a/ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs
+++ b/ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs
@@ -218,6 +218,7 @@ namespace ExtractorUtils.Test.Unit.Unstable
             await checkIn.Flush(source.Token);
             Assert.Equal(2, _checkInCallbacks.Count);
             Assert.Equal(2, _checkInCallbacks[1]);
+            Assert.Equal(1, _checkInCallbacks[0]);
         }
 
 
--- a/ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs
+++ b/ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs
@@ -61,6 +61,8 @@ namespace ExtractorUtils.Test.Unit.Unstable
             };
         }
 
+        private List<int> _checkInCallbacks = new();
+
         private (ServiceProvider, CheckInWorker) GetCheckInWorker()
         {
             var config = GetConfig();
@@ -77,7 +79,13 @@ namespace ExtractorUtils.Test.Unit.Unstable
 
             var client = provider.GetRequiredService<Client>();
 
-            return (provider, new CheckInWorker(config.Integration, provider.GetRequiredService<ILogger<CheckInWorker>>(), client));
+            return (provider, new CheckInWorker(
+                config.Integration,
+                provider.GetRequiredService<ILogger<CheckInWorker>>(),
+                client,
+                _checkInCallbacks.Add,
+                0
+            ));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
```diff
@@ -223,5 +255,27 @@ public void ReportTaskStart(string taskName, TaskUpdatePayload? update = null, D
...
+            {
+                if (_onRevisionChanged != null)
+                {
+                    _logger.LogInformation("Remote config revision changed {From} -> {To}", _activeRevision, response.LastConfigRevision);
+                    _onRevisionChanged(response.LastConfigRevision.Value);
+                }
+                else
+                {
+                    _logger.LogInformation(
+                        "Remote config revision changed {From} -> {To}. The extractor is currently using local configuration and will not restart.",
```
ozangoktan, line 273: Beyond the possible confusion with local(file) config vs remote config, wouldn't it be better here to log that the extractor is not setup to automatically follow remote config updates.

ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs
```diff
@@ -175,6 +186,40 @@ public async Task TestCheckInWorkerBatch()
...
+            // Flush without changes
+            await checkIn.Flush(source.Token);
+            Assert.Single(_checkInCallbacks);
+            Assert.Equal(1, _checkInCallbacks[0]);
+
+            // Flush with a new revision
+            _lastConfigRevision = 2;
+            await checkIn.Flush(source.Token);
+            Assert.Equal(2, _checkInCallbacks.Count);
+            Assert.Equal(2, _checkInCallbacks[1]);
```
ozangoktan, line 220:
```suggestion
            Assert.Equal(2, _checkInCallbacks[1]);
            Assert.Equal(1, _checkInCallbacks[0]);
```

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
einarmo: Isn't that what we're logging though? Isn't it better to log _why_ we're not following config updates?

einarmo: Yeah, we can add that, but we should mention that it's because it's running using local config. That message is _less_ informative than the current one.

einarmo: It's also wrong. Restarting the extractor won't do anything unless you configure it to use remote config, even if you manually restart it.

ozangoktan: Is it correct to refer to the running config as a local config? I see it more as the current config than a local config. Reusing the likeness of the static local file config does not seem right to me when we are using a set version of the remote config. 

Another thing is if there is 2 changes in the remote, then this would be saying the remote jumped 2 versions, which is incorrect as well.

ozangoktan: Oh, I think I confused this with an extractor that has remote config but requires manual restarts to switch configs. Is that a usecase we are planning for?

einarmo: Also, if the user updated the config twice and it jumped two revisions, that _would_ be correct. Why would it be wrong?

einarmo: I'm just going to remove the check entirely and always call the callback, I think that's better anyway.

ozangoktan (approved):

PR #531, opened 2025-03-28
Report startup

einarmo:
Final part of startup reporting. This ensures that we actually send the startup request before we report _any_ checkins, thus avoiding inconsistent state in CDF, even if the extractor runs offline for some time before being able to report startup.

The ability to run offline is opt-in, so the checkin worker can be configured to crash if we cannot connect to CDF.

### DIFF
--- a/ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
+++ b/ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
@@ -90,27 +90,49 @@ namespace Cognite.Extractor.Utils.Unstable.Tasks
             // Make sure the external ID in the startup payload matches the external ID of the target integration.
             startupPayload.ExternalId = _integrationId;
 
-            while (!token.IsCancellationRequested)
+            // Hold the flush lock while reporting startup, to ensure that we don't start reporting checkins
+            // before the startup request has been sent.
+            // With this, calls to flush will wait until the startup request has been sent,
+            // or startup fails.
+            // This prevents unfortunate behavior if we recover connection to CDF, then immediately shut down.
+            // In this case, we would like to potentially report a startup and anything that has happened
+            // while the connection to CDF was down.
+            try
             {
-                try
-                {
-                    await ReportStartup(startupPayload, token).ConfigureAwait(false);
-                    _hasReportedStartup = true;
-                    break;
-                }
-                catch (Exception ex)
-                {
-                    if (!_retryStartup) throw;
-                    var toDelay = _random.Next(STARTUP_BACKOFF_SECONDS / 2, STARTUP_BACKOFF_SECONDS * 3 / 2);
-                    _logger.LogError("Failed to report startup, retrying after {Time} seconds: {Err}", toDelay, ex.Message);
-                    await Task.Delay(TimeSpan.FromSeconds(toDelay), token).ConfigureAwait(false);
-                    continue;
-                }
+                await _flushLock.WaitAsync(token).ConfigureAwait(false);
+            }
+            catch
+            {
+                // Should only happen if we are cancelled while waiting.
+                if (token.IsCancellationRequested) return;
+                throw;
             }
 
-            lock (_lock)
+            try
             {
-                _hasReportedStartup = true;
+                while (!token.IsCancellationRequested)
+                {
+                    try
+                    {
+                        await ReportStartup(startupPayload, token).ConfigureAwait(false);
+                        _hasReportedStartup = true;
+                        break;
+                    }
+                    catch (Exception ex)
+                    {
+                        if (!_retryStartup) throw;
+                        // Retry about every 30 seconds, but with some jitter so we don't
+                        // end up with too bursty retries.
+                        var toDelay = _random.Next(STARTUP_BACKOFF_SECONDS / 2, STARTUP_BACKOFF_SECONDS * 3 / 2);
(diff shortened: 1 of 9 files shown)


ExtractorUtils/Unstable/Tasks/Errors.cs
```diff
@@ -187,9 +187,11 @@ public interface IIntegrationSink
         /// Run the sink, automatically flushing at regular intervals.
         /// </summary>
         /// <param name="token">Cancellation token.</param>
+        /// <param name="startupPayload">Payload to send to the startup endpoint before beginning to
+        /// report periodic checkins..</param>
```
ozangoktan, line 191:
```suggestion
        /// report periodic checkins.</param>
```

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
```diff
@@ -41,19 +48,24 @@ public class CheckInWorker : IIntegrationSink
         /// <param name="activeRevision">Currently active config revision. Used to know whether the extractor has received a new
         /// config revision since the last checkin. Null indiciates that the extractor is running local config,
         /// and should not restart based on changes to remote config.</param>
+        /// <param name="retryStartup">Whether to retry the startup request if it fails,
+        /// beyond normal retries. If this is `true`, the checkin worker will retry startup requests forever,
+        /// or until they succeed.</param>
```
ozangoktan, line 53:
```suggestion
        /// beyond normal retries. If this is `true`, the check-in worker will retry startup requests indefinitely,
        /// instead of raising an exception.</param>
```

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
```diff
@@ -62,15 +74,67 @@ public CheckInWorker(
         /// This may only be called once.
         /// </summary>
         /// <param name="token">Cancellation token</param>
+        /// <param name="startupPayload">Payload to send to the startup endpoint before beginning to
+        /// report periodic checkins..</param>
         /// <param name="interval">Interval, defaults to 30 seconds.</param>
-        public async Task RunPeriodicCheckin(CancellationToken token, TimeSpan? interval = null)
+        public async Task RunPeriodicCheckin(CancellationToken token, StartupRequest startupPayload, TimeSpan? interval = null)
```
ozangoktan, line 80: CheckIn? I also see a bunch of docstrings/comments that say "checkins" rather than "check-ins", from this PR and from before.

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
```diff
@@ -62,15 +74,67 @@ public CheckInWorker(
...
             }
 
+            // Make sure the external ID in the startup payload matches the external ID of the target integration.
+            startupPayload.ExternalId = _integrationId;
+
+            // Hold the flush lock while reporting startup, to ensure that we don't start reporting checkins
+            // before the startup request has been sent.
+            // With this, calls to flush will wait until the startup request has been sent,
+            // or startup fails.
+            // This prevents unfortunate behavior if we recover connection to CDF, then immediately shut down.
```
ozangoktan, line 97:
```suggestion
            // This keeps us from reporting events before the startup, in case the extractor is started offline.
```

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
```diff
@@ -98,6 +162,27 @@ public async Task RunPeriodicCheckin(CancellationToken token, TimeSpan? interval
...
+            catch
+            {
+                // Only really happens if we're cancelled while waiting.
+                return;
+            }
+
+            if (!_hasReportedStartup)
+            {
+                _logger.LogWarning("Attempted to flush checkin worker before reporting startup, since this results in unclear errors in CDF, the request is ignored. It would likely fail.");
+                return;
```
ozangoktan, line 183: Doesn't this leave the flush lock unreleased?

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
einarmo: We can change that later, but this is the wrong PR for it.

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
einarmo: It does indeed, good catch.

ExtractorUtils/Unstable/Tasks/Task.cs
```diff
@@ -24,6 +24,34 @@ public TaskUpdatePayload(string? message = null)
...
+    /// Task metadata object.
+    /// </summary>
+    public class TaskMetadata
+    {
+        /// <summary>
+        /// Type of task.
+        /// </summary>
+        public TaskType Type { get; set; }
+        /// <summary>
+        /// Whether the task can be triggered by an action or not.
```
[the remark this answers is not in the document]
einarmo: I still want it to be part of the API for now, since it will be needed eventually.

ExtractorUtils/Unstable/Tasks/CheckInWorker.cs
einarmo: Never mind, I think we should change it. I'll just do it here...

ozangoktan (approved):

psalaberria002 (approved):
:unicorn:

PR #538, opened 2025-04-10
Various things for final runtime implementation

einarmo:
The runtime is too big to fit in under 500 lines, this adds some disjointed things that are necessary for it. I couldn't find a way to split the runtime method itself, which is way too complex, so I split of everything else.

That does mean parts of this PR makes little sense, since there is no context to be had.

### DIFF
--- a/ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
+++ b/ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
@@ -4,11 +4,24 @@ using System.Threading.Tasks;
 using Cognite.Extractor.Testing;
 using Cognite.Extractor.Utils.Unstable.Runtime;
 using Cognite.Extractor.Utils.Unstable.Configuration;
+using Cognite.Extractor.Configuration;
 using ExtractorUtils.Test.Unit.Unstable;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Abstractions;
+using System.Net.Http;
+using System;
+using System.Net;
+using Newtonsoft.Json;
+using System.Collections.Generic;
+using CogniteSdk.Alpha;
+using System.Dynamic;
+using System.Net.Http.Headers;
+using Cognite.Extractor.Utils.Unstable.Tasks;
+using CogniteSdk;
+using Cognite.Extractor.Utils.Unstable;
+using Cognite.Extensions.Unstable;
 
 namespace ExtractorUtils.Test.unit.Unstable
 {
@@ -49,7 +62,7 @@ authentication:
             var configPath = TestUtils.AlphaNumericPrefix("dotnet_extractor_test") + "_config";
             builder.ConfigFolder = configPath;
             Directory.CreateDirectory(configPath);
-            File.WriteAllText(configPath + "/connection.yml", GetConnectionConfig());
+            System.IO.File.WriteAllText(configPath + "/connection.yml", GetConnectionConfig());
 
             var runtime = await builder.MakeRuntime(CancellationToken.None);
 
@@ -71,7 +84,7 @@ authentication:
             var configPath = TestUtils.AlphaNumericPrefix("dotnet_extractor_test") + "_config";
             builder.ConfigFolder = configPath;
             Directory.CreateDirectory(configPath);
-            File.WriteAllText(configPath + "/connection.yml", GetConnectionConfig());
+            System.IO.File.WriteAllText(configPath + "/connection.yml", GetConnectionConfig());
 
             var runtime = await builder.MakeRuntime(CancellationToken.None);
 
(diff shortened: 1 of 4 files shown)


ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -97,5 +110,162 @@ public async Task TestRuntimeBuilderStatic()
...
+        }
+
+        private List<dynamic> taskEvents = new();
+        private List<dynamic> errors = new();
+        private List<dynamic> startupRequests = new();
+        private int _checkInCount;
+        private int _startupCount;
+        private ConfigRevision _responseRevision;
+
+        private async Task<HttpResponseMessage> mockRequestsAsync(
```
ozangoktan, line 178: This method is similar enough to what we have in ExtractorUtils.Test/unit/Unstable/CheckInWorkerTests.cs that we may want to think about having it in a shared place.

einarmo: I agree, I think a shared CDF mock may be worthwhile, but I want to make a larger effort, not a tiny part of something in a different PR.

ExtractorUtils/Unstable/Runtime/Bootstrap.cs
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    _logger.LogWarning("No integration provided for error reporting. Errors will not be sent to CDF.");
+                }
+
+                _pendingErrors.Clear();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError("Failed to report errors to CDF: {Message}", ex.Message);
+            }
```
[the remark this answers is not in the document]
einarmo: IIRC I only call flush once on this type. But I can do the whole song and dance of pulling them out, then re-adding them if the flush fails, just to be consistent.

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -97,5 +110,162 @@ public async Task TestRuntimeBuilderStatic()
...
+                Authentication = new ClientCredentialsConfig
+                {
+                    ClientId = "someId",
+                    ClientSecret = "thisIsASecret",
+                    Scopes = new Cognite.Common.ListOrSpaceSeparated("https://greenfield.cognitedata.com/.default"),
+                    MinTtl = "60s",
+                    Resource = "resource",
+                    Audience = "audience",
+                    TokenUrl = "http://example.url/token",
+                }
```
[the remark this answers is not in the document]
einarmo: This is all dummy data, and they are clearly different, since one is serialized. There's generalization, which is fine, but I feel like trying to unify these two cases will just make the test code harder to reason about.

ozangoktan (dismissed):

ExtractorUtils/Unstable/Runtime/Bootstrap.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Cognite.Extractor.Common;
+using Cognite.Extractor.Utils.Unstable.Tasks;
+using CogniteSdk;
+using CogniteSdk.Alpha;
+using Microsoft.Extensions.Logging;
+
+namespace Cognite.Extractor.Utils.Unstable.Runtime
```
frxstrem, line 12: This file contains two separate classes, should they possibly be in two separate files?

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -97,5 +110,162 @@ public async Task TestRuntimeBuilderStatic()
...
+            };
+        }
+
+        [Fact(Timeout = 5000)]
+        public async Task TestBootstrapSinks()
```
frxstrem, line 135: This seems to test two separate classes sequentially, maybe it should be split into two separate tests?

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -97,5 +110,162 @@ public async Task TestRuntimeBuilderStatic()
...
+            services.AddCogniteClient("myApp");
+            services.AddConfig(GetConfig(), typeof(ConnectionConfig));
+            var provider = services.BuildServiceProvider();
+            var sink2 = new BootstrapErrorReporter(
+                provider.GetRequiredService<Client>(),
+                "test-integration",
+                provider.GetRequiredService<ILogger<RuntimeTest>>()
+            );
+            using var e2 = new ExtractorError(ErrorLevel.error, "test", sink2, "details", null, DateTime.UtcNow);
+            e.Instant();
```
frxstrem, line 158: Should this be `e2.Instant();`?

einarmo: Yup, good catch...

frxstrem (commented):
:unicorn:

PR #539, opened 2025-04-15
Add an optional timeout to the periodic scheduler

einarmo:
This can cause extractor shutdown to hang forever if there's an issue with shutting down a task, so we should allow setting a timeout.

Should we set a default? It would technically be a breaking change, but hanging forever seems unlikely to be desirable behavior.


thorkildcognite:
risk review ok

PR #540, opened 2025-04-23
Add auth using middleware instead of through the SDK

einarmo:
By adding auth through middleware instead of on top of the client, we should be able to automatically retry 401s, which could avoid hard errors when the token expires while retrying.

This makes it so that:
  - We retry requests in the authenticator.
  - We get a new token if it expires before we retry a request to CDF.

Which should really help with avoiding spurious errors.

The reason why this is done here instead of in the SDK is that the SDK is fundamentally built to leave configuring the http client to the user, and this must be done before the http client is created.

Since we're in the middle of rewriting the library (in the unstable module), there are some duplicated changes in this PR.


ExtractorUtils.Test/unit/CogniteTest.cs
```diff
@@ -150,6 +150,60 @@ public async Task TestClientRetry()
...
+                .ConfigureCogniteHttpClientHandlers();
+
+            services.AddCogniteClient("testApp", setLogger: true, setMetrics: true, setHttpClient: false);
+            using var provider = services.BuildServiceProvider();
+            var cogniteDestination = provider.GetRequiredService<CogniteDestination>();
+            await cogniteDestination.CogniteClient.Assets.ListAsync(new AssetQuery());
+
+            // First we hit the auth endpoint 3 times, to get an initial valid token.
+            // Next, we hit the assets endpoint once, it fails, and when retrying we need to fetch a new token.
+            // This means we hit the auth endpoint 6 times in total.
```
ozangoktan, line 202: Would it make sense to test out a configuration where we do not have enough retries configured? To nail down that behavior as well.

finnag (approved):
:unicorn:

PR #541, opened 2025-04-25
Make checking the project optional

einarmo:
It's nice to have an endpoint we can call to check if CDF is actually available at all, but currently this means that extractors require groups:list and projects:list, which isn't ideal. This makes that check optional.

It is enabled by default for backwards compatibility.


ozangoktan (approved):

larshagencognite (approved):
🦄

PR #543, opened 2025-04-30
Add field to extractor to set the scheduler timeout

einarmo:
There's currently no way to pass this to the base extractor, so the timeout feature isn't very helpful...


ozangoktan (approved):

larshagencognite (approved):
🦄

PR #549, opened 2025-06-10
Avoid upserting spaces unnecessarily in tests

einarmo:
This was causing schema lock conflicts and failing the tests. Instead just get the space ID when needed.

### DIFF
--- a/ExtractorUtils.Test/CDFTester.cs
+++ b/ExtractorUtils.Test/CDFTester.cs
@@ -66,6 +66,7 @@ namespace ExtractorUtils.Test
             }
 
             await DestinationWithIDM.CogniteClient.DataModels.UpsertSpaces(new List<SpaceCreate>() { new() { Space = _spaceId } });
+            _spaceUpserted = true;
 
             return _spaceId;
         }
--- a/ExtractorUtils.Test/CDFTester.cs
+++ b/ExtractorUtils.Test/CDFTester.cs
@@ -32,7 +32,9 @@ namespace ExtractorUtils.Test
         public string Prefix { get; private set; }
         public BaseConfig Config { get; }
         private readonly string _configPath;
-        public string SpaceId { get; private set; }
+
+        private string _spaceId;
+        private bool _spaceUpserted;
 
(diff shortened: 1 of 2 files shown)


ozangoktan (approved):

ExtractorUtils.Test/CDFTester.cs
```diff
@@ -32,7 +32,9 @@ class CDFTester : IDisposable
         public string Prefix { get; private set; }
         public BaseConfig Config { get; }
         private readonly string _configPath;
-        public string SpaceId { get; private set; }
+
+        private string _spaceId;
+        private bool _spaceUpserted;
```
finnag, line 37: I agree with the check warning here - this is never assigned to, is this intentional?

finnag (changes requested):
check warning that should be looked into

ExtractorUtils.Test/CDFTester.cs
einarmo: Indeed... Setting that is the entire point of the PR.

ExtractorUtils.Test/CDFTester.cs
```diff
@@ -50,11 +52,23 @@ public CDFTester(string[] config, ITestOutputHelper output)
...
             Source = new CancellationTokenSource();
-            SpaceId = $"dotnet-extractor-utils-test-space{i}";
+            _spaceId = $"dotnet-extractor-utils-test-space{i}";
         }
         public CDFTester(CogniteHost host, ITestOutputHelper output) : this(GetConfig(host), output)
         {
-            DestinationWithIDM.CogniteClient.DataModels.UpsertSpaces(new List<SpaceCreate>() { new() { Space = SpaceId } }).GetAwaiter().GetResult();
+        }
+
+        public async Task<string> GetSpaceId()
```
larshagencognite, line 61: should this be synchronized to avoid concurrent loads? Not a big deal since only test code

larshagencognite (approved):
🦄

finnag (approved):

PR #554, opened 2025-06-23
feat: introduce CogniteExtractorTimeSeriesBase for adding an option t…

121yaseen:
…o the opcua extractor to store the output as JSON


einarmo (commented):
You will want to change the `where` bounds in https://github.com/cognitedata/dotnet-extractor-utils/blob/master/Cognite.Extensions/DataModels/CogniteExtractorExtensions/ExtractorTimeSeriesExtensions.cs, I think, so that we can still use those methods.

The methods will need to be generic over one more parameter, but I think that's possible.

121yaseen:
> You will want to change the `where` bounds in https://github.com/cognitedata/dotnet-extractor-utils/blob/master/Cognite.Extensions/DataModels/CogniteExtractorExtensions/ExtractorTimeSeriesExtensions.cs, I think, so that we can still use those methods.
> 
> The methods will need to be generic over one more parameter, but I think that's possible.
> 
> Also, please update the PR title and description.

Along with that, won't we require to update the users of the methods in the extensions as well?

PR #555, opened 2025-06-25
Align connection config with python utils

einarmo:
This is discussed in the config design doc. Move things around a bit to better align with the layout we agreed on for python utils.

This removes the SDK logging and makes it enabled by default on debug level, only in the unstable module for now.

### DIFF
--- a/ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
+++ b/ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
@@ -41,7 +41,8 @@ namespace ExtractorUtils.Test.unit.Unstable
             return @"
 project: project
 base-url: https://api.cognitedata.com
-integration: test-integration
+integration:
+    external-id: test-integration
 authentication:
   type: client-credentials
   client-id: someId
@@ -346,7 +347,10 @@ authentication:
             {
                 Project = "project",
                 BaseUrl = "https://greenfield.cognitedata.com",
-                Integration = "test-integration",
+                Integration = new IntegrationConfig
+                {
+                    ExternalId = "test-integration"
+                },
                 Authentication = new ClientCredentialsConfig
                 {
                     ClientId = "someId",
--- a/ExtractorUtils/Unstable/Configuration/Connection.cs
+++ b/ExtractorUtils/Unstable/Configuration/Connection.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using Cognite.Extensions.Unstable;
+using Cognite.Extractor.Common;
 using Cognite.Extractor.Configuration;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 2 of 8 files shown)


ExtractorUtils/Unstable/Configuration/Connection.cs
```diff
@@ -31,24 +32,14 @@ public class ConnectionConfig : VersionedConfig
...
-        public string? Integration { get; set; }
+        public IntegrationConfig? Integration { get; set; }
 
         /// <summary>
-        /// Configuration for retries of failed requests to CDF.
+        /// Configuration for the connection to CDF.
         /// </summary>
-        public RetryConfig CdfRetries { get => _cdfRetries; set { _cdfRetries = value ?? _cdfRetries; } }
-        private RetryConfig _cdfRetries = new RetryConfig();
+        public CdfConnectionConfig CdfConnection { get => _cdfConnection; set { _cdfConnection = value ?? _cdfConnection; } }
```
polomani, line 40: why `_cdfConnection = value ?? _cdfConnection;` here?
it won't allow resetting the value. is it intentional?

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -41,7 +41,8 @@ private string GetConnectionConfig()
             return @"
 project: project
 base-url: https://api.cognitedata.com
-integration: test-integration
+integration:
+    external-id: test-integration
```
polomani, line 45: This change is breaking, but I presume it's an experimental API?

polomani:
Left some questions to clarify

ExtractorUtils/Unstable/Configuration/Connection.cs
einarmo: It won't allow setting the value to null, which is intentional. It's the only real way to have a non-nullable property with yaml, that I have found.

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
einarmo: Yep. Everything under Unstable is, reasonably enough, experimental.

ExtractorUtils/Unstable/DestinationUtilsUnstable.cs
```diff
@@ -110,9 +110,9 @@ public static IHttpClientBuilder ConfigureCogniteHttpClientHandlers(this IHttpCl
             {
                 try
                 {
-                    var retryConfig = provider.GetService<ConnectionConfig>()?.CdfRetries;
+                    var retryConfig = provider.GetService<ConnectionConfig>()?.CdfConnection.Retries ?? new Configuration.RetryConfig();
```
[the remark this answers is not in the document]
einarmo: I don't understand why the coverage claims these lines aren't covered. We do test this in `ConnectionConfigTests.cs`, and the coverage I see in codecov doesn't make any sense. The `try` is covered, but neither the `catch` or the actual body of the try are covered? That doesn't seem possible. This code should run unconditionally when this method is called, and the stuff under `ConfigurePrimaryHttpMessageHandler` _is_ covered. I have no idea how I'd make sure this code is covered beyond what I've already done. I think it is a bug in the coverage collector.

PR #556, opened 2025-06-26
make base class to enable writing as JSON to timeseries

121yaseen:
bump up the version to add base class to enable writing as JSON to timeseries


einarmo (approved):
:unicorn:

PR #557, opened 2025-07-07
Set StartTime in Startup request

einarmo:
This is necessary for sequencing if the extractor is disconnected from CDF when starting. Otherwise, we would end up with conflicts if any tasks were closed by the startup request.


toondaey (approved):

einarmo:
:unicorn:

PR #563, opened 2025-08-01
CDF mocking framework

einarmo:
We have various ways to do this in different extractors, and it would be nice to have a common framework. The plan is to extend the tests in the utils to also use this.

There's some unpleasant DI magic to get this to work properly with the extensions we add to HTTP clients, but it works fairly well, and is hopefully flexible enough that it can be expanded for greater ease of use. In theory we could even create pre-built mocks for common CDF APIs.


einarmo:
:unicorn: - big PR, but this is all test utilities, so the impact is relatively low. It's also self-tested.

PR #565, opened 2025-08-05
Use bearer token auth in the authenticator

einarmo:
CogIdp doesn't support sending the client secret as a form-encoded parameter. This is the preferred method anyway.

We keep the old way of doing it, to avoid regressions.

Also improve logging a bit, potentially extracting request ID.

### DIFF
--- a/Cognite.Extensions/Authenticator.cs
+++ b/Cognite.Extensions/Authenticator.cs
@@ -344,9 +344,9 @@ namespace Cognite.Extensions
             using var request = new HttpRequestMessage(HttpMethod.Post, _config.TokenUrl);
             request.Content = httpContent;
             request.Headers.Add("Accept", "application/json");
-            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", $"{_config.ClientId}:{_config.ClientSecret}");
+            request.Headers.Authorization = new AuthenticationHeaderValue("Basic", Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes($"{_config.ClientId}:{_config.ClientSecret}")));
 
-            var response = await _client.SendAsync(request, token).ConfigureAwait(false);
+            using var response = await _client.SendAsync(request, token).ConfigureAwait(false);
 #if NET5_0_OR_GREATER
             var body = await response.Content.ReadAsStringAsync(token);
 #else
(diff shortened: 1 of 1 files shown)


Cognite.Extensions/Authenticator.cs
```diff
@@ -338,51 +340,72 @@ private async Task<ResponseDTO> RequestToken(CancellationToken token = default)
...
             var form = _config.GetFormdata();
 
-            using (var httpContent = new FormUrlEncodedContent(form))
-            {
-                var response = await _client.PostAsync(_config.TokenUrl, httpContent, token);
+            using var httpContent = new FormUrlEncodedContent(form);
+            using var request = new HttpRequestMessage(HttpMethod.Post, _config.TokenUrl);
+            request.Content = httpContent;
+            request.Headers.Add("Accept", "application/json");
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", $"{_config.ClientId}:{_config.ClientSecret}");
```
121yaseen, line 347: If its clientId and Secret, shouldn't this be Basic?

Cognite.Extensions/Authenticator.cs
```diff
@@ -338,51 +340,72 @@ private async Task<ResponseDTO> RequestToken(CancellationToken token = default)
...
+
+            var response = await _client.SendAsync(request, token).ConfigureAwait(false);
```
121yaseen, line 349: Shouldn't `response` also need to be disposed?

121yaseen:
Sorry, meant to do as a single review, came out a 3 individual comments

Cognite.Extensions/Authenticator.cs
einarmo: I'm not sure why the compiler isn't complaining about this, we usually get warnings.

Cognite.Extensions/Authenticator.cs
```diff
@@ -338,51 +340,72 @@ private async Task<ResponseDTO> RequestToken(CancellationToken token = default)
...
+                }
+
+                _logger.LogDebug(
+                    "New OIDC token. Expires on {ttl}",
+                    (DateTime.UtcNow + TimeSpan.FromSeconds(tokenResponse.ExpiresIn)).ToISOString());
+                return tokenResponse;
+            }
+            else
+            {
+                var requestId = response.Headers.TryGetValues("X-Request-ID", out var reqId) ? reqId.FirstOrDefault() : null;
```
121yaseen, line 375: Here, assuming that header will always be `X-Request-ID` and not `X-Request-Id`

121yaseen (approved):

Cognite.Extensions/Authenticator.cs
einarmo: HTTP headers are case-insensitive, according to the standard.

PR #574, opened 2025-08-20
Unstable fixes

einarmo:
Four minor fixes to the unstable module:

 - Treat InfiniteTimeSpan as infinite when scheduling tasks.
 - Pass the correct cancellation token to `InitConfigSource`, this was causing the extractor to hang if you tried to cancel it while it was failing due to missing connection config or the like.
 - Truncate extractor version. While extractor versions aren't longer than 32 characters, for local development or dev builds they can sometimes include revision info. The limit on the API is 32 chars.
 - Properly handle task scheduler cancellation. When the scheduler shuts down we do want to report that tasks have ended. I also changed it so we report a warning here, not a fatal failure, since that seems a bit strong. We should somehow report that the task didn't properly shut down, however.

I have a pair of larger fixes as well, that will come in separate PRs.

### DIFF
--- a/ExtractorUtils.Test/unit/Unstable/TaskSchedulerTest.cs
+++ b/ExtractorUtils.Test/unit/Unstable/TaskSchedulerTest.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


ozangoktan (approved):

einarmo:
:unicorn:

PR #583, opened 2025-09-16
Migrate CogniteTest to new HTTP mock framework

einarmo:
This serves to verify the implementation, and check that we actually get proper retries working, which we generally don't with other forms of mocking.

I also add a few utilities, and improve the code in some places.

I think the actual diff is negative, but the added utilities and general code pushes it to about equal. Still a win, IMO. The code is generally easier to read and work with.

### DIFF
--- a/ExtractorUtils.Test/unit/Unstable/TaskSchedulerTest.cs
+++ b/ExtractorUtils.Test/unit/Unstable/TaskSchedulerTest.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Cognite.Extractor.Common;
+using Cognite.Extractor.Testing;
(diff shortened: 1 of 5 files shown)


Cognite.Testing/Mock/BuiltIns.cs
```diff
@@ -21,7 +23,7 @@ public static void AddTokenEndpoint(this CdfMock mock, int minRequests = 0, int
                 tokenResponse.token_type = "Bearer";
                 tokenResponse.expires_in = 3600;
                 tokenResponse.access_token = "test-access-token";
-                return ctx.CreateJsonResponse(tokenResponse);
+                return (HttpResponseMessage)ctx.CreateJsonResponse(tokenResponse);
```
121yaseen, line 26: Isn't the explicit cast to (HttpResponseMessage) is redundant? As CreateJsonResponse already returns this type

121yaseen (approved):
Minor Comment. Everything else seems good.

PR #584, opened 2025-09-17
Add initial (low noise) gemini configuration

thorkildcognite:
We are rolling out Gemini to all risk reviewed repositories, but before then we want to have a basic low-risk configuration rolled out


thorkildcognite:
risk review ok

PR #590, opened 2025-10-06
Pair of fixes for unstable

einarmo:
- Set the cancellation token source to null after disposing it.
 - Avoid weird .NET race condition.

The .NET race condition is a _terrifying_ footgun that I wasn't aware of. If you call `SetResult` on a `TaskCompletionSource` the continuation (i.e. the next block of synchronous code after where the task is awaited), runs _immediately_. Since the caller here holds a lock, and .NET locks are (apparently) re-entrant, this means that this implicitly bypasses the lock, allowing separate pieces of code to modify the content at the same time.


einarmo:
:unicorn:

PR #591, opened 2025-10-07
Use CdfMock for DatapointsTest

einarmo:
This adds mocks for timeseries byids and insert datapoints to the common mocks, and replaces some of the tests in `DatapointsTest.cs`. Due to line limit I didn't get to the delete datapoints test, but it should be reasonably simple to write.

### DIFF
--- a/ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
+++ b/ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
@@ -127,6 +127,7 @@ namespace ExtractorUtils.Test.Integration
             try
             {
                 var result = await tester.DestinationWithIDM.EnsureTimeSeriesExistsAsync(timeseries, RetryMode.OnError, SanitationMode.Clean, tester.Source.Token);
+                tester.Logger.LogResult(result, RequestType.UpsertInstances, false);
                 created = result.Results;
 
                 Assert.Single(result.Errors);
@@ -162,6 +163,7 @@ namespace ExtractorUtils.Test.Integration
             try
             {
                 var result = await tester.DestinationWithIDM.UpsertTimeSeriesAsync(updates, RetryMode.None, SanitationMode.None, tester.Source.Token);
+                tester.Logger.LogResult(result, RequestType.UpsertInstances, false);
                 result.Throw();
 
                 Assert.Equal(3, result.Results.Count());
@@ -205,6 +207,7 @@ namespace ExtractorUtils.Test.Integration
             try
             {
                 var result = await tester.DestinationWithIDM.UpsertTimeSeriesAsync(updates, RetryMode.OnError, SanitationMode.Clean, tester.Source.Token);
+                tester.Logger.LogResult(result, RequestType.UpsertInstances, false);
                 result.ThrowOnFatal();
 
                 Assert.Equal(2, result.Results.Count());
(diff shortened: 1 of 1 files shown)


Cognite.Testing/Mock/TimeSeries.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// <param name="id">Timeseries ID.</param>
+        /// <param name="isString">Indicates if the timeseries is of string type.</param>
+        public void MockTimeSeries(Identity id, bool isString)
+        {
+            if (id == null) throw new ArgumentNullException(nameof(id));
+            var ts = new TimeSeries
+            {
+                Id = id.Id ?? Interlocked.Increment(ref _nextId),
+                ExternalId = id.ExternalId,
+                InstanceId = id.InstanceId != null ? new InstanceIdentifier
```
ozangoktan: Why are we re-creating this? Can't we just assign?

Cognite.Testing/Mock/TimeSeries.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// <param name="timeSeries">Timeseries metadata.</param>
+        public void MockTimeSeries(Identity id, TimeSeries timeSeries)
+        {
+            if (id == null) throw new ArgumentNullException(nameof(id));
+            if (timeSeries == null) throw new ArgumentNullException(nameof(timeSeries));
+            if (timeSeries.Id == 0)
+            {
+                timeSeries.Id = Interlocked.Increment(ref _nextId);
+            }
+            _timeSeries[id] = new TimeSeriesWrapper(timeSeries);
```
ozangoktan: Seems a bit weird to me, we are taking 2 id fields, but not enforcing them to be consistent. Shouldn't we at least update the internal id's to be consistent before returning?

Cognite.Testing/Mock/TimeSeries.cs
```diff
@@ -0,0 +1,337 @@
...
+        public RequestMatcher GetByIdsMatcher(Times times)
+        {
+            return new SimpleMatcher("POST", "/timeseries/byids", MockTimeSeriesByIds, times);
+        }
+
+        /// <summary>
+        /// Get a matcher for the /timeseries/data endpoint.
+        /// </summary>
+        /// <param name="times">Expected number of executions.</param>
+        public RequestMatcher CreateDatapointsMatcher(Times times)
```
ozangoktan: Shouldn't this be GetDatapointsMatcher as to align with the one above? In general it may be good to avoid using "Create" when naming there, as to avoid actually confusing it with the one that calls "POST" on "/timeseries".

Cognite.Testing/Mock/TimeSeries.cs
```diff
@@ -0,0 +1,337 @@
...
+                {
+                    Code = 400,
+                    Message = $"Expected {mismatchedExpected} value for datapoint",
+                });
+            }
+
+            foreach (var item in data.Items)
+            {
+                var id = GetIdentity(item);
+                var ts = _timeSeries[id];
```
ozangoktan: Maybe we should keep the timeseries references we fetched in the missing check around, as to avoid doing this, as this could create thread safety issues once delete feature comes around.

ozangoktan:
> Of course once I add logging to the IDM tests they pass. We need to do something about them, either nail down what the eventual consistency in the API is, or change them somehow.

I agree, maybe create some discovery task? I've yet to see a consistent pattern we can use as a success criteria.

PR #654, opened 2026-04-08
add Add integration test for GetExtractedRanges

ozangoktan:
This adds an integration test for GetExtractedRanges which tests both ignoreBadDataPoints true and false behavior. Follow-up from #646


nithinb (approved):

polomani:
🦄

PR #655, opened 2026-04-09
Restart extractor on config change

Toshad:
Currently, we restart on config change only if restart policy is 'always' but we also want to restart for other cases to support integrations.


ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -415,7 +429,11 @@ private async Task<ExtractorRunResult> BuildAndRunExtractor(ServiceProvider prov
...
                 return ExtractorRunResult.Error;
             }
-
+            if (shouldRestart)
+            {
+                _activeLogger.LogInformation("Extractor stopped cleanly with policy {Policy}, restart is required", _params.RestartPolicy);
+                return ExtractorRunResult.RestartRequired;
+            }
             return ExtractorRunResult.CleanShutdown;
         }
```
Hmnt39, line 440: Can we add one more test which test config change should still restart, even with Never policy?

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -147,6 +153,11 @@ public async Task Run()
                     // Otherwise, immediately restart.
                     backoff = 0;
                 }
+                else if (result == ExtractorRunResult.RestartRequired)
+                {
+                    _activeLogger.LogInformation("Extractor stopped cleanly with restart required, restarting with backoff");
+                    backoff = 1;
```
Hmnt39, line 159: Should this backoff need to be configurable? backoff(1) means 5ms?

Hmnt39 (commented):
Is this restart trigger will only be done on remote config changes or local remote will also do it ?

ExtractorUtils/Unstable/Runtime/Runtime.cs
Toshad: We don't want to restart with Never policy. Never should mean never.

ExtractorUtils/Unstable/Runtime/Runtime.cs
Toshad: It means 5s. ~If we want to make it configurable, I think we should do it separately, that isn't relevant to this change.~
It is already configurable.

Toshad:
> Is this restart trigger will only be done on remote config changes or local remote will also do it ?

If connection config is set to local, then it will do it.

If connection config is set to 'no connection', Then it will not do it.

I think this is correct behaviour.

ExtractorUtils/Unstable/Runtime/Runtime.cs
Hmnt39: but, here it is hardcoded to 1?

Toshad: This is how backoff is treated in the entire file, as an int which is later multiplied by `BackoffBase`.
Like I said earlier, this is not the right diff to change that behaviour anyway.

ExtractorUtils/Unstable/Runtime/Runtime.cs
Toshad: Added a test to ensure that behaviour.

Hmnt39 (dismissed):

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -147,6 +153,11 @@ public async Task Run()
                     // Otherwise, immediately restart.
                     backoff = 0;
                 }
+                else if (result == ExtractorRunResult.RestartRequired)
```
aasmundeldhuset, line 156: Not something that needs to be addressed in this PR, but: shouldn't this large if/else over an enum be a `switch` statement?

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -36,6 +36,11 @@ enum ExtractorRunResult
         /// The extractor crashed.
         /// </summary>
         Error,
+        /// <summary>
+        /// The extractor was stopped with a clean shutdown.
+        /// But we need to restart it (possibly due to a revision change).
+        /// </summary>
+        RestartRequired
```
aasmundeldhuset: Nitpick that you don't have to address, but: adding a comma at the end will prevent an unnecessary modification of this line the next time an enum member is added.

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -328,6 +339,7 @@ private async Task<ExtractorRunResult> BuildAndRunExtractor(ServiceProvider prov
         {
             using var internalTokenSource = CancellationTokenSource.CreateLinkedTokenSource(_source.Token);
             TExtractor extractor;
+            var shouldRestart = false;
```
aasmundeldhuset: This isn't used in the first `try`, so I think it should be declared above the second one, so that its limited scope is more clear.

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -415,7 +429,11 @@ private async Task<ExtractorRunResult> BuildAndRunExtractor(ServiceProvider prov
 
                 return ExtractorRunResult.Error;
             }
-
+            if (shouldRestart)
```
aasmundeldhuset, line 433: Add a blank line above to visually differentiate this from the try/catch.

On a related note: this is a 100-line function that has got three quite distinct parts: the creation of the extractor, the waiting for the extractor or the "revision" change event, and the final restart check. I recommend splitting out the first two into separate functions - but that doesn't need to be done in this PR.

aasmundeldhuset (dismissed):
:unicorn: 

I only have style feedback; the two that are in scope of this PR are small enough that you can ignore them if you want - but I strongly recommend breaking down the large function to make it easier to see the high-level flow (in a separate PR).

ExtractorUtils/Unstable/Runtime/Runtime.cs
Toshad: Will raise a separate PR

Hmnt39 (dismissed):

Toshad:
@aasmundeldhuset I made some of the improvements you suggested, pls re-approve.

ExtractorUtils/Unstable/Runtime/Runtime.cs
```diff
@@ -110,42 +115,44 @@ public async Task Run()
...
                         backoff = 0;
-                    }
-                }
-                else if (result == ExtractorRunResult.CleanShutdown)
-                {
-                    // Shut down, if the extractor is configured to only restart on error.
-                    if (_params.RestartPolicy != ExtractorRestartPolicy.Always)
-                    {
-                        _activeLogger.LogInformation("Extractor closed cleanly, shutting down");
                         break;
```
aasmundeldhuset, line 151: Oh no - I forgot that C# uses `break` to mark the end of each `switch` block. So the change from `if` to `switch` causes the `break` on line 147 to change meaning: it no longer stops the `while` loop; it just jumps out of the `switch`. I guess you'll have to change back to `if` - apologies. However, it's quite concerning that no tests broke because of this. And that reminds me of something I should have pointed out yesterday: there should be a test that verifies that the extractor actually restarts when the run result is `RestartRequired`. Please write that test.

Toshad: @aasmundeldhuset We already have that test, which is the one I modified.
This error doesn't affect it because this `if` only executes for `CleanShutdown`, which doesn't seem to have a test. Will add it now.

aasmundeldhuset:
Ready to risk approve when the lint has been fixed.

Hmnt39 (approved):

aasmundeldhuset (commented):
:unicorn:

PR #658, opened 2026-04-16
Allow option to add reason to task cancellations

Toshad:
When we cancel tasks currently, we don't allow extractor to pass any
details of why this task was cancelled. This diff allows passing those details


ExtractorUtils.Test/unit/Unstable/TaskSchedulerTest.cs
```diff
@@ -289,5 +289,51 @@ public async Task TestSchedulerCancel()
...
+
+            sched.AddScheduledTask(task, true);
+            startEvt.WaitOne();
+
+            // Task should be canceled while waiting for it to finish, and WaitForNextEndOfTask should throw TaskCanceledException.
+            var waitTask = sched.WaitForNextEndOfTask("Task1", TimeSpan.FromSeconds(3));
+
+            sched.CancelTask("Task1", "Test cancellation");
+
+            await Assert.ThrowsAsync<AggregateException>(async () => await waitTask);
```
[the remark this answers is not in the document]
Toshad: Test doesn't fail as exceptions are aggregated.

ExtractorUtils/Unstable/Tasks/ExtractorTaskScheduler.cs
```diff
@@ -146,13 +147,14 @@ public void Run(DateTime now, CancellationToken token)
...
 
-        public void Cancel()
+        public void Cancel(string? reason = null)
         {
             lock (_lock)
             {
                 if (ActiveTask != null)
                 {
                     ActiveTask.Source.Cancel();
+                    ActiveTask.CancellationReason = reason;
```
Hmnt39, line 157:
```suggestion
                    ActiveTask.CancellationReason = reason;
                    ActiveTask.Source.Cancel();
```
Reason should be set first before cancellation

ExtractorUtils/Unstable/Tasks/ExtractorTaskScheduler.cs
```diff
@@ -518,7 +521,7 @@ private async Task RunInner(CancellationToken token)
                     if (minNextRun != null)
                     {
 #pragma warning disable CA2016 // Forward the 'CancellationToken' parameter to methods
-                        toAwait.Add(Task.Delay(tickTime - minNextRun.Value));
+                        toAwait.Add(Task.Delay(minNextRun.Value - tickTime));
```
Hmnt39, line 525: Is `Task.delay` able to handle negative value? If not, then maximum should be the first value

Hmnt39 (changes requested):

ExtractorUtils/Unstable/Tasks/ExtractorTaskScheduler.cs
Toshad: `minNextRun.Value` will be greater than `tickTime`, see the `if else` at line 501.

Toshad: Added a test around this anyway.

Hmnt39 (approved):
Looks good to me

MyBitBird (commented):
🦄

PR #659, opened 2026-04-17
Don't fail fatally for all runtime errors

Toshad:
When an extractor fails during its run, we catch the exception and log a Fatal , but don't rethrow the exception.
This is incorrect behaviour and this diff fixes it.


ExtractorUtils/Unstable/BaseExtractor.cs
Toshad, line 248: It is unclear to me if the original intention was to:
1. Catch exceptions thrown during ordinary task runs, or
2. If this was just for exceptions from `TaskScheduler` and `RunPeriodicCheckIn`.

Another question remains, should the extractor restart when `TaskScheduler`/`PeriodicCheckIn` throw exception?

ExtractorUtils/Unstable/BaseExtractor.cs
```diff
@@ -249,14 +249,23 @@ public async Task Start(CancellationToken token)
...
-                return;
+                if (ex is ConfigurationException)
+                {
+                    // We shouldn't ever reach here since config exceptions should be caught in Init and reported as startup errors,
+                    // but just in case, we want to make sure these are treated as fatal and reported to the user as they are actionable.
+                    // If there's a genuine case of fatal errors seen during runtime, we should consider introducing a new Exception for that case,
+                    // and add it here, to avoid accidentally treating transient errors as fatal.
+                    Fatal(flattened.Message, flattened.StackTrace?.ToString());
+                    return;
+                }
```
ozangoktan: This is new functionality where we make ConfigurationException into a magic exception that does not allow restarts when thrown from anywhere in any extractor if it is allowed to hit the utils level. We can discuss adding it, but let's not do that in this PR, especially without tests around it.

ExtractorUtils/Unstable/Tasks/ExtractorTaskScheduler.cs
```diff
@@ -453,6 +453,7 @@ public async Task<SchedulerTaskResult> Run(CancellationToken token)
                 }
                 else
                 {
+                    _logger.LogError(ex, "Scheduler failed with exception: {Message}", ex.Message);
```
ozangoktan: This should already be logged [here](https://github.com/cognitedata/dotnet-extractor-utils/blob/baa67b375b7203a8730bc3529bd154c71592ce30/ExtractorUtils/Unstable/Runtime/Runtime.cs#L408) with the line just below this setting the exception at the task level.

ExtractorUtils.Test/unit/Unstable/BaseExtractorTest.cs
```diff
@@ -158,7 +158,8 @@ public async Task TestBaseExtractorMonitoredError()
                 throw new Exception("Monitored error");
             }, SchedulerTaskResult.Unexpected, "task1");
             var delayTask = Task.Delay(2000);
-            Assert.NotEqual(delayTask, await Task.WhenAny(runTask, delayTask));
+            var ranTask = await Task.WhenAny(runTask, delayTask);
+            Assert.NotEqual(delayTask, ranTask);
```
ozangoktan: Why not this?
```suggestion
            Assert.Equal(runTask, ranTask);
```

ozangoktan (commented):
Quite confused why this was returning rather than throwing. If I had context back then, I cannot recall it, and looking at the structure, I can't find why this would be done so.

The change from return to throw makes sense to me, and not creating a Fatal error for any exception also makes sense.

There are some extra changes in the PR that I don't agree with. Can you take a look at the comments?

Also, can you add a test that actually checks if the extractor restarts? In RuntimeTest.cs, I see no test checking for _startupCount > 1 that deals with runtime errors.

ozangoktan (approved):
LGTM

ExtractorUtils.Test/unit/Unstable/RuntimeTest.cs
```diff
@@ -337,6 +337,7 @@ public async Task TestRuntimeRestartBadConfig()
 
             builder.MaxBackoff = 100;
             builder.BackoffBase = 10;
+            builder.RestartPolicy = ExtractorRestartPolicy.OnError;
```
Toshad, line 340: Note for reviewer:
This test could've caught the issue if this line had existed, so added it. In general when Restart is being tested, the default `ExtractorRestartPolicy.Always` shouldn't be used.

polomani (commented):
🦄

PR #662, opened 2026-05-08
DM TimeSeries Type Mismatch Handling

ozangoktan:
Add functionality for handling CDM Timeseries type mismatch when working with views inheriting CDM Timeseries


Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,71 @@
...
+                        var (cleanItems, itemsWithTypeError) = CleanTypeImmutabilityError(items, fetched);
+                        if (itemsWithTypeError.Count > 0)
+                        {
+                            error.Values = itemsWithTypeError.Select(x => new Identity(new InstanceIdentifier(x.Space, x.ExternalId))).Concat(error.Values ?? new List<Identity>());
+                            error.Skipped = itemsWithTypeError.Concat(error.Skipped ?? new List<SourcedNodeWrite<T>>());
+                        }
+                        return cleanItems;
+                    }
+                }
+                // else if(typeof(CogniteAssetBase).IsAssignableFrom(typeof(T))) { }
```
Toshad, line 67: nit: Remove this if not needed.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,71 @@
...
+            IEnumerable<SourcedNodeWrite<T>> items,
+            T2 resource,
+            CancellationToken token
+            ) where T2 : BaseDataModelResource<T>
+        {
+            if (resource != null && error?.Type == ErrorType.IllegalItem)
+            {
+                if (typeof(CogniteTimeSeriesBase).IsAssignableFrom(typeof(T)))
+                {
+                    var fetched = await FetchItems(resource, items, token).ConfigureAwait(false);
```
Toshad: THis should be inside the `if`, we're not using it otherwise.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/ExtractorTimeSeriesResultHandlers.cs
```diff
@@ -1,34 +1,64 @@
...
+            foreach (var item in items)
+            {
+                var identifier = new InstanceIdentifier(item.Space, item.ExternalId);
+                if (!foundTypeDict.ContainsKey(identifier) || foundTypeDict[identifier] == (item.Properties as CogniteTimeSeriesBase)?.Type)
+                {
+                    cleanItems.Add(item);
+                }
+                else
+                {
+                    skipped.Add(item);
```
Toshad, line 57: nit: Maybe add a debug log line here.

Toshad:
`CleanFromErrorCommon` seems to return an empty list when none of the items are affected, but return cleaned items if one of them is affected. Is this correct? 
Moreover, the returned items list isn't used by callers, the callers only care about the errors. Is this expected behavior? Why is it designed in this awkward way?

Cognite.Extensions/DataModels/CogniteExtractorExtensions/ExtractorTimeSeriesResultHandlers.cs
ozangoktan: There is no logger here, this stuff would be logged upstream. Would be too verbose to put in debug level anyway.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: If I remove this, I should collapse the 2 nested if blocks also. This structure is meant as a guideline for when we are developing a handler for some other resource type, and this comment somewhat serves as an explanation.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: We are using it, just that it is declared here for it to be used when we write other handlers for the same type. Added a lazy load mechanism for this https://github.com/cognitedata/dotnet-extractor-utils/pull/662/commits/952dd016814bdd821b7522b23e069e84bbc2b0c5

ozangoktan:
> `CleanFromErrorCommon` seems to return an empty list when none of the items are affected, but return cleaned items if one of them is affected. Is this correct? Moreover, the returned items list isn't used by callers, the callers only care about the errors. Is this expected behavior? Why is it designed in this awkward way?

Interesting question, this code pre-dates my time working on this codebase, so, we can discuss what this method is doing and why it is structured as such, but I don't think this discussion should have an effect with regards to reviewing this PR. The new code is there so that this method is not called when we can handle the case better.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
Toshad: Could you add a text comment clarifying this?

Toshad (dismissed):

Toshad:
> Interesting question, this code pre-dates my time working on this codebase, so, we can discuss what this method is doing and why it is structured as such, but I don't think this discussion should have an effect with regards to reviewing this PR. The new code is there so that this method is not called when we can handle the case better.

Yes, I'd like to know why this method is structured as such since this has come up.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,73 @@
...
+                        var (cleanItems, itemsWithTypeError) = CleanTypeImmutabilityError(items, await FetchItemsCached().ConfigureAwait(false));
+                        if (itemsWithTypeError.Count > 0)
+                        {
+                            error.Values = itemsWithTypeError.Select(x => new Identity(new InstanceIdentifier(x.Space, x.ExternalId))).Concat(error.Values ?? new List<Identity>());
+                            error.Skipped = itemsWithTypeError.Concat(error.Skipped ?? new List<SourcedNodeWrite<T>>());
+                        }
+                        return cleanItems;
+                    }
+                }
+                // else if(typeof(CogniteAssetBase).IsAssignableFrom(typeof(T))) { }
```
Toshad, line 67:
```suggestion
                // Other handlers can later be added similarly
                // else if(typeof(CogniteAssetBase).IsAssignableFrom(typeof(T))) { }
```

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                if (typeof(CogniteTimeSeriesBase).IsAssignableFrom(typeof(T)))
+                {
+                    IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>? __fetched = null;
+                    Func<Task<IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>>> FetchItemsCached = async () => { __fetched ??= await FetchItems(resource, items, token).ConfigureAwait(false); return __fetched; };
+
+                    if (error.Message?.Contains("'cdf_cdm.CogniteTimeSeries.type'") == true)
+                    {
+                        var (cleanItems, itemsWithTypeError) = CleanTypeImmutabilityError(items, await FetchItemsCached().ConfigureAwait(false));
+                        if (itemsWithTypeError.Count > 0)
```
Toshad, line 58: Is it possible for `itemsWithTypeError` to be empty, despite `ErrorType.IllegalItem`? If yes, doesn't this have a totally different behavior from `CleanFromErrorCommon` ?

ozangoktan: This is an edge case where the upsert fails due to type mismatch but someone deletes the offending timeseries right before we fetch them with CDM TimeSeries view. Can you clarify what you mean about the behavior difference, and why it has to be identical?

Toshad: The behavior doesn't _have_ to be identical, but since we're calling `CleanFromErrorCommon` in the default case here, it would be better.

The difference in behavior is the case where all items are clean:
1. If the type is `CogniteTimeSeriesBase`, we return the clean items.
2. Else we return empty iterable(from `CleanFromErrorCommon`)

ozangoktan: I don't get your perspective here, and I don't think 1 is correct at all. cleanItems reflects items that are clean, therefore ready for use. If the problem is gone while we are handling it, it is valid to return the items that are clean. If our specific handler is not a perfect fit for the issue, we let the common handler run.

Toshad: Yeah, I guess it makes sense when put it like that.

Toshad (dismissed):

Cognite.Extensions/CogniteResult.cs
```diff
@@ -121,6 +121,10 @@ public static CogniteError<TError> ParseException<TError>(Exception ex, RequestT
                 {
                     ParseDatapointsException(rex, result);
                 }
+                else if (type == RequestType.UpsertInstances)
```
aasmundeldhuset: Wouldn't this entire thing be better as a `switch`? (Doesn't need to be handled in this PR; just a suggestion.)

Cognite.Extensions/DataModels/QueryBuilder/Filter.cs
```diff
@@ -232,7 +232,7 @@ public static IDMSFilter HasData(IEnumerable<SourceIdentifier> sources)
             if (sources == null) throw new ArgumentNullException(nameof(sources));
             return new HasDataFilter
             {
-                Models = sources
```
aasmundeldhuset, line 235: How did this work before? Is `models` an undocumented filter name in DMS that is the same as `hasData`?

ozangoktan: Ref: https://github.com/cognitedata/cognite-sdk-dotnet/pull/471

Cognite.Extensions/CogniteResult.cs
ozangoktan: I agree, I could just drop the change in if @Toshad agrees also, otherwise we could create a simple task for it. Should be fairly simple to do either way.

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
```diff
@@ -42,18 +39,34 @@ private static SourcedNodeWrite<CogniteExtractorTimeSeries> GetWritableTS(CDFTes
...
-                GetWritableTS(tester, "utils-test-ts-1", spaceId, extraPopulateMethod: x => {x.Properties.Name = "utils-test-ts-1"; return x;}),
-                GetWritableTS(tester, "utils-test-ts-2", spaceId, TimeSeriesType.String, x => {x.Properties.Name = "utils-test-ts-2"; return x;}),
-                GetWritableTS(tester, "utils-test-ts-3", spaceId, extraPopulateMethod: x => {x.Properties.Name = "utils-test-ts-3"; return x;}),
-            };
+            var timeseries = Enumerable.Range(offset + 1, count).Select(i => GetWritableTS<T>(
+                tester,
+                $"{tester.Prefix}utils-test-ts-{i}",
+                spaceId,
+                i % 2 == 0 ? TimeSeriesType.String : TimeSeriesType.Numeric,
+                x => { x.Properties.Name = $"utils-test-ts-{i}"; return x; }
```
aasmundeldhuset, line 50: Should the prefix have been used here too?

aasmundeldhuset, line 50: Should the prefix have been used here too?

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
```diff
@@ -221,30 +234,30 @@ public async Task TestUpdateSanitation(CogniteHost host)
             }
             finally
             {
-                await tester.DestinationWithIDM.CogniteClient.DataModels.DeleteInstances(tss.Select(x => new InstanceIdentifierWithType(InstanceType.node, x)), tester.Source.Token);
+                await DeleteTimeseries(tester, spaceId, tss.externalIds);
             }
         }
 
-        // This feature depends on fields currently not present on DM, ex: https://api-docs.cognite.com/20230101/tag/Time-series/operation/alterTimeSeries#!c=422&path=error/duplicated&t=response
```
aasmundeldhuset, line 228: Shouldn't there instead be a leading space after each `//`?

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
ozangoktan: We don't add the prefix to the name property in the object, don't see a reason to.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="items">Nodes to clean</param>
+        /// <param name="resource">Resource to query for underlying timeseries with, when T is a child of CogniteTimeSeries</param>
+        /// <param name="token">Cancellation token</param>
+        /// <returns>Nodes that are not affected by the error</returns>
+        public static async Task<IEnumerable<SourcedNodeWrite<T>>> CleanFromError<T, T2>(
+            CogniteError<SourcedNodeWrite<T>> error,
+            IEnumerable<SourcedNodeWrite<T>> items,
+            T2 resource,
+            CancellationToken token
+            ) where T2 : BaseDataModelResource<T>
```
aasmundeldhuset: Dedent one level (to match `public`) unless a formatter requires this level, to get a visual difference between the parameter list and generic type constraints.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            return error.Resource switch
+            {
+                ResourceType.InstanceId => badValues.ContainsIdentity(new InstanceIdentifier(ts.Space, ts.ExternalId)),
+                _ => false,
+            };
+        }
+
+        /// <summary>
+        /// Clean list of SourceNodeWrite objects based on CogniteError
```
aasmundeldhuset: Maybe say something about what "clean" means here?

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+        public static async Task<IEnumerable<SourcedNodeWrite<T>>> CleanFromError<T, T2>(
+            CogniteError<SourcedNodeWrite<T>> error,
+            IEnumerable<SourcedNodeWrite<T>> items,
+            T2 resource,
+            CancellationToken token
+            ) where T2 : BaseDataModelResource<T>
+        {
+            if (resource != null && error?.Type == ErrorType.IllegalItem)
+            {
+                if (typeof(CogniteTimeSeriesBase).IsAssignableFrom(typeof(T)))
```
aasmundeldhuset, line 50: The comment below suggests that this can be expanded in the future to handle more types. Is there a `switch` or `match` construct in C# that is better suited for matching on type?

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>? __fetched = null;
```
aasmundeldhuset, line 52: I know about the underscore convention in C# for member properties, but is there really a convention about double underscores too? What does it signify?

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+                    Func<Task<IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>>> FetchItemsCached = async () => { __fetched ??= await FetchItems(resource, items, token).ConfigureAwait(false); return __fetched; };
+
+                    if (error.Message?.Contains("'cdf_cdm.CogniteTimeSeries.type'") == true)
```
aasmundeldhuset, line 55: Are the single quotes here intentional?

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
ozangoktan: I would assume so myself, but given it was Visual Studio's built in block comment functionality that formatted it this way, I am inclined to say this is the way to do it.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
aasmundeldhuset: (Is it because it ends up in a lambda closure?)

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
```diff
@@ -0,0 +1,74 @@
...
+            T2 resource,
+            CancellationToken token
+            ) where T2 : BaseDataModelResource<T>
+        {
+            if (resource != null && error?.Type == ErrorType.IllegalItem)
+            {
+                if (typeof(CogniteTimeSeriesBase).IsAssignableFrom(typeof(T)))
+                {
+                    IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>? __fetched = null;
+                    Func<Task<IEnumerable<SourcedInstance<CogniteTimeSeriesBase>>>> FetchItemsCached = async () => { __fetched ??= await FetchItems(resource, items, token).ConfigureAwait(false); return __fetched; };
```
aasmundeldhuset, line 53: Why make `__fetched` contain the results if you're accessing them via the function's return value? Seems to me this could just be a boolean. (But I'm very outdated on C#, so maybe this is an established async idiom?)

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: Yes, there are single quotes in the error message.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
aasmundeldhuset: Er, never mind, I now see how it becomes a memoized/cached function.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: Single underscore signifies field/property that is private. Double underscore here is me trying to signify "please don't use this by value". I would appreciate a suggestion if you have any.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: I don't think we can reasonably switch-case from this conditional, it is checking if the type inherits a specific base class.

Cognite.Extensions/CogniteResult.cs
ozangoktan: Did the change, see https://github.com/cognitedata/dotnet-extractor-utils/pull/662/commits/396ecb0c37a2b078720435bc803a10830097016b

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
ozangoktan: At least added the space back on the comment. https://github.com/cognitedata/dotnet-extractor-utils/pull/662/commits/396ecb0c37a2b078720435bc803a10830097016b

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
aasmundeldhuset: Looks like you can: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns (I would assume the cases don't require exact matches, just is-subtype-of).

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
aasmundeldhuset: The thing you're doing seems like a reusable pattern for caching an async function. Even if you're currently only using it here, maybe you could extract it to a function that takes the actual fetch function as a parameter and returns the cached function? Then the variable is hidden inside the function, and it can be reused elsewhere.

ozangoktan: Given this is static context, I don't think that works. The cached value is local scoped, and we cannot use static scope because it wouldn't be thread-safe.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
ozangoktan: But we are doing inheritability check on the generic type itself, not an instance of it. I don't think that example applies.

aasmundeldhuset: Oh, right, you don't actually have an instance of type `T`.

Cognite.Extensions/DataModels/InstancesResultHandlers.cs
aasmundeldhuset: Then I have no further suggestions :sweat_smile:

aasmundeldhuset (commented):
:unicorn: 

Thanks for the in-person walkthrough! However, one more question for consideration: if `FetchItemsCached` fails (e.g. because of a 429 rate limiting), what error message will be bubbled up to the user? If the error message contains specific about that particular operation, might that cause confusion since the user didn't directly request that lookup?

ozangoktan:
> Thanks for the in-person walkthrough! However, one more question for consideration: if `FetchItemsCached` fails (e.g. because of a 429 rate limiting), what error message will be bubbled up to the user? If the error message contains specific about that particular operation, might that cause confusion since the user didn't directly request that lookup?

We do have retries and jitter, and if everything fails, we would bubble up an exception. It is not acceptable to continue, as the error would just keep happening otherwise.

Toshad (approved):

PR #663, opened 2026-05-19
Add support to report errors as 'config' type to integrations

Toshad:
Currently, we report all errors as general type, which causes an issue when config read fails,
because the errors are reported against the previous active configs. Integrations API has 
'config' error type just for this use case, so this diff enables using it.

It is dependent on corresponding changes in SDK at https://github.com/cognitedata/cognite-sdk-dotnet/pull/485.

Manual test result:
Screenshot 2026-05-19 at 3 42 39 PM

Also added a line to ensure we log what log config was derived. This is useful to debug in cases where log isn't getting printed, and extractor config can from different places.


ExtractorUtils/Configuration/RemoteConfig.cs
```diff
@@ -168,7 +168,7 @@ private T InjectRemoteConfig(T config)
             {
                 var rawConfig = await _destination.CogniteClient.ExtPipes.GetCurrentConfigAsync(_pipelineId, token).ConfigureAwait(false);
                 var config = ConfigurationUtils.TryReadConfigFromString<T>(rawConfig.Config, _acceptedConfigVersions);
-
+                _logger.LogCritical("Successfully fetched configuration from CDF for pipeline {Pipeline}. Revision: {Rev}", _pipelineId, rawConfig.Revision);
```
vikramlc-cognite: Question: Shouldn't this be LogInformation?

Cognite.Logging/Logger.cs
```diff
@@ -124,6 +124,8 @@ public static Serilog.ILogger GetConfiguredLogger(LoggerConfig config)
         public static Serilog.ILogger GetConfiguredLogger(LoggerConfig config, LogEventLevel baseMinLevel)
         {
             var logConfig = GetConfiguration(config, baseMinLevel);
+            GetSerilogDefault().Information("Logger configured with minimum level {Level} and output to Console: {Console}, File: {File}, Stderr: {Stderr}",
```
vikramlc-cognite:
```suggestion
            GetSerilogDefault().Information("Configuring Logger with minimum level {Level} and output to Console: {Console}, File: {File}, Stderr: {Stderr}",
```

ExtractorUtils/Unstable/Tasks/Errors.cs
```diff
@@ -30,14 +30,22 @@ public class ExtractorError : IDisposable
...
         /// </summary>
         public string? Details { get; set; }
+        /// <summary>
+        /// Optional error type, used to categorize errors.
+        /// </summary>
+        public string? Type { get; set; }
+        /// <summary>
+        /// Config revision active when this error was reported.
+        /// </summary>
+        public int? ConfigRevision { get; set; }
```
vikramlc-cognite:
```suggestion
        public string? Type { get; }
        public int? ConfigRevision { get; }
```

ExtractorUtils/Unstable/Tasks/Errors.cs
```diff
@@ -268,10 +288,12 @@ public void Error(string description, string? details = null, DateTime? now = nu
...
         /// <param name="details">Long fatal error details.</param>
         /// <param name="now">Current time, for synchronization. Defaults to DateTime.UtcNow</param>
+        /// <param name="type">Type of the error.</param>
+        /// <param name="configRevision">Configuration revision associated with the error.</param>
         /// <returns>An error that should be completed later.</returns>
-        public ExtractorError BeginFatal(string description, string? details = null, DateTime? now = null)
+        public ExtractorError BeginFatal(string description, string? details = null, DateTime? now = null, string? type = null, int? configRevision = null)
         {
-            return NewError(ErrorLevel.fatal, description, details, now);
+            return NewError(ErrorLevel.fatal, description, details, now, type, configRevision);
```
vikramlc-cognite, line 296: Question: Can we add **type** and **configRevision** also to **BeginWarning** and **BeginError** ?

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
```diff
@@ -155,7 +155,7 @@ public override async Task<bool> ResolveConfig(int? targetRevision, BaseErrorRep
                 var msg = ex.Message;
                 if (_lastErrorMsg != msg)
                 {
-                    reporter.Fatal($"Fatally failed to load configuration file from {_configFilePath}: {ex.Message}");
+                    reporter.Fatal($"Fatally failed to load local configuration file from {_configFilePath}: {ex.Message}", type: "config");
```
vikramlc-cognite, line 158: Question: Should we also pass ex.ToString() as details to the log as well which will have more details rather than just exception message?

ExtractorUtils/Unstable/Tasks/Errors.cs
Toshad: I think we should keep it as is until integrations api allows some other type of errors, we want all config errors to be fatal.

ExtractorUtils/Unstable/Configuration/ConfigSource.cs
Toshad: It is not the convention followed in this repo, but especially for config errors I don't think it makes sense.
The message already has what line in config failed to parse, and why. The stacktrace, etc wouldn't be useful for anyone IMO.

vikramlc-cognite (dismissed):
Approved

MyBitBird (commented):
Looks good to me, but the Gemini's comment seems valid, please check it out and also have a test to cover it.

Toshad:
@MyBitBird Fixed the gemini comment.

vikramlc-cognite (approved):

MyBitBird:
@Toshad CI is failing

Toshad:
@MyBitBird It was some flaky error in creating the docker image for redis-github-action. Retry succeeded.

MyBitBird:
🦄

PR #665, opened 2026-05-22
Add empty and null check for external_id when fetching ranges

nithinb:
When we check for externalId value, we only check if its null. In the cases where there is an instanceId is defined, externalId defaults to an empty string, which matches the condition of `dp.ExternalId != null` and this would cause it to execute the if-block instead of the `else if (dp.InstanceId != null)` block.

The changes made here are quite simple, 

* Change the check from `!= null` to `!string.IsNullOrEmpty(dp.ExternalId)`
* Add tests for this case
* Bump the version of utils


ozangoktan (commented):
The external ids being checked for emptiness is coming from the timeseries API. If they are empty, I think there is no graceful way of handling anything.

ozangoktan:
> The external ids being checked for emptiness is coming from the timeseries API. If they are empty, I think there is no graceful way of handling anything.

Looking at your PR description now, I misunderstood in my initial comment, and asking around a bit, pretty surprised to learn that empty external ids are a thing in the classic api! I would appreciate an integration test recreating this, actually.

nithinb:
@ozangoktan I actually wanted to add an integration test in the extractor itself but yea, can add that here itself.

ozangoktan:
> @ozangoktan I actually wanted to add an integration test in the extractor itself but yea, can add that here itself.

@nithinb The integration test passes even when I revert the null-or-empty changes. I was hoping for an integration test that actually recreates this.

nithinb:
@ozangoktan it fails for me if I undo the fix.


```
  ExtractorUtils.Test test net8.0 failed with 2 error(s) (11.9s)
    /Users/nithin.bodanapu@cognitedata.com/Documents/work/repos/dotnet-extractor-utils/ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs(565): error TESTERROR: 
      ExtractorUtils.Test.Integration.IDMTimeSeriesIntegrationTest.TestGetExtractedRangesInstanceIdResolution(host: BlueField) (10s 998ms): Error Message:
       Assert.Equal() Failure: Strings differ
                 ↓ (pos 0)
      Expected: "2026-05-22 10:43:11.806"
      Actual:   "1970-01-01 00:00:00.000"
                 ↑ (pos 0)

```

Screenshot 2026-05-22 at 16 14 11

ozangoktan:
> @ozangoktan it fails for me if I undo the fix.
> 
> ```
>   ExtractorUtils.Test test net8.0 failed with 2 error(s) (11.9s)
>     /Users/nithin.bodanapu@cognitedata.com/Documents/work/repos/dotnet-extractor-utils/ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs(565): error TESTERROR: 
>       ExtractorUtils.Test.Integration.IDMTimeSeriesIntegrationTest.TestGetExtractedRangesInstanceIdResolution(host: BlueField) (10s 998ms): Error Message:
>        Assert.Equal() Failure: Strings differ
>                  ↓ (pos 0)
>       Expected: "2026-05-22 10:43:11.806"
>       Actual:   "1970-01-01 00:00:00.000"
>                  ↑ (pos 0)
> ```
> 

Oh, nevermind, it was my test setup being broken somehow.

ExtractorUtils.Test/integration/IDMTimeSeriesIntegrationTest.cs
```diff
@@ -522,5 +521,57 @@ public async Task TestDataPointsCreateMissing(CogniteHost host)
...
+                foreach (var id in identities)
+                {
+                    Assert.True(result.ContainsKey(id), $"Missing range for {id}");
+                    tester.Logger.LogInformation(
+                        "GetExtractedRanges [{id}] — First: {first}, Last: {last} (expected First: {expectedFirst})",
+                        id, result[id].First.ToISOString(), result[id].Last.ToISOString(), firstTimestamp.ToISOString());
+                    // First must be the actual earliest datapoint, not DateTimeEpoch.
+                    // With the old dp.ExternalId != null check, proto responses return ExternalId=""
+                    // which is non-null, causing identity resolution to fall through to Identity(dp.Id)
+                    // (not in idSet) — so First stays at DateTimeEpoch and this assertion fails.
```
ozangoktan: This comment will be entirely irrelevant past this PR.

nithinb: oh yea that is fair. Thanks

ozangoktan (dismissed):

ozangoktan (commented):
Version needs to be updated, it is no longer valid.

thorkildcognite:
@nithinb Can you please fix the conflicts?

Toshad (approved):
Pls check if this change needed elsewhere as well

polomani (commented):
🦄

PR #674, opened 2026-07-01
Add support for State Sets

Toshad:
CDF now supports a first-class CogniteTimeSeries.type = "state" that stores both a numericValue and a stringValue on every data point and links the time series to a CogniteStateSet node that catalogs the valid states. 

This PR adds support to allow downstream extractors to create CogniteStateSet, and adds generic helper functions to CRUD beta resources

### DIFF
--- a/Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
+++ b/Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
@@ -1,0 +1,236 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Cognite.Extractor.Common;
+using CogniteSdk;
+using CogniteSdk.DataModels;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Prometheus;
+
+namespace Cognite.Extensions.DataModels.CogniteExtractorExtensions
+{
+    /// <summary>
+    /// Retrieves instances by id for the beta CDM resource wrapped by a <see cref="BetaResourceExtensions"/> method.
+    /// </summary>
+    internal delegate Task<IEnumerable<SourcedNode<T>>> RetrieveInstancesFunc<T>(IEnumerable<InstanceIdentifierWithType> ids, CancellationToken token);
+
+    /// <summary>
+    /// Upserts instances for the beta CDM resource wrapped by a <see cref="BetaResourceExtensions"/> method.
+    /// </summary>
+    internal delegate Task<IEnumerable<SlimInstance>> UpsertInstancesFunc<T>(IEnumerable<SourcedNodeWrite<T>> items, CancellationToken token);
+
+    /// <summary>
+    /// Sanitizes a batch of instance writes for the beta CDM resource wrapped by a <see cref="BetaResourceExtensions"/> method.
+    /// </summary>
+    internal delegate (IEnumerable<SourcedNodeWrite<T>>, IEnumerable<CogniteError<SourcedNodeWrite<T>>>) SanitizeInstancesFunc<T>(IEnumerable<SourcedNodeWrite<T>> items, SanitationMode mode);
+
+    /// <summary>
+    /// Shared get-or-create implementation for beta CDM resources
+    /// (<see cref="CogniteSdk.Resources.Beta.StateSetsResource"/>, <see cref="CogniteSdk.Resources.Beta.TimeSeriesResource"/>).
+    /// These resources do not implement <c>BaseDataModelResource&lt;T&gt;</c>, so they cannot use the
+    /// generic implementation in <see cref="Cognite.Extensions.DataModels.DataModelUtils"/>; instead
+    /// <see cref="BetaStateSetsExtensions"/> is a thin public wrapper that supplies its resource's
+    /// retrieve/upsert/sanitize operations to the methods here.
+    /// </summary>
+    internal static class BetaResourceExtensions
+    {
+        private static ILogger _logger = new NullLogger<Client>();
+
+        internal static void SetLogger(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public static Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateAsync<T>(
+            ViewIdentifier view,
+            RetrieveInstancesFunc<T> retrieve,
+            UpsertInstancesFunc<T> upsert,
+            SanitizeInstancesFunc<T> sanitize,
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, IEnumerable<SourcedNodeWrite<T>>> buildItems,
+            BetaResourceParams options,
+            CancellationToken token)
+        {
+            Task<IEnumerable<SourcedNodeWrite<T>>> AsyncBuildItems(IEnumerable<InstanceIdentifier> ids)
+            {
+                return Task.FromResult(buildItems(ids));
+            }
+            return GetOrCreateAsync(view, retrieve, upsert, sanitize, instanceIds, AsyncBuildItems, options, token);
+        }
+
+        public static async Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateAsync<T>(
+            ViewIdentifier view,
+            RetrieveInstancesFunc<T> retrieve,
+            UpsertInstancesFunc<T> upsert,
+            SanitizeInstancesFunc<T> sanitize,
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, Task<IEnumerable<SourcedNodeWrite<T>>>> buildItems,
+            BetaResourceParams options,
+            CancellationToken token)
+        {
+            var chunks = instanceIds
+                .ChunkBy(options.ChunkSize)
+                .ToList();
+            if (!chunks.Any()) return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(null, null);
+
+            var results = new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>[chunks.Count];
+
+            var generators = chunks
+                .Select<IEnumerable<InstanceIdentifier>, Func<Task>>(
+                    (chunk, idx) => async () =>
+                    {
+                        results[idx] = await GetOrCreateChunk(view, retrieve, upsert, sanitize, chunk,
+                            buildItems, 0, options, token).ConfigureAwait(false);
+                    });
+
+            await generators.RunThrottled(options.ThrottleSize, token).ConfigureAwait(false);
+
+            return CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>.Merge(results);
+        }
+
+        private static async Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateChunk<T>(
+            ViewIdentifier view,
+            RetrieveInstancesFunc<T> retrieve,
+            UpsertInstancesFunc<T> upsert,
+            SanitizeInstancesFunc<T> sanitize,
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, Task<IEnumerable<SourcedNodeWrite<T>>>> buildItems,
+            int backoff,
+            BetaResourceParams options,
+            CancellationToken token)
+        {
+            IEnumerable<SourcedNode<T>> found;
+            using (CdfMetrics.Instances(view, "retrieve").NewTimer())
+            {
+                try
+                {
+                    found = await retrieve(
+                        instanceIds.Select(x => new InstanceIdentifierWithType(InstanceType.node, x)), token).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    var err = ResultHandlers.ParseSimpleError<SourcedNodeWrite<T>>(ex, instanceIds?.Select(x => Identity.Create(x)), null);
+                    return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(new[] { err }, null);
+                }
+            }
+
+            var missing = instanceIds.Except(found.Select(ts => new InstanceIdentifier(ts.Space, ts.ExternalId))).ToList();
+
+            if (missing.Count == 0)
+            {
+                return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(null, found);
+            }
+
+            var toCreate = await buildItems(missing).ConfigureAwait(false);
+
+            IEnumerable<CogniteError<SourcedNodeWrite<T>>> errors;
+            (toCreate, errors) = sanitize(toCreate, options.SanitationMode);
+
+            var result = await CreateHandleErrors(view, upsert, toCreate, options.RetryMode, token).ConfigureAwait(false);
+            result.Results = result.Results == null ? found : result.Results.Concat(found);
+
+            if (errors.Any())
+            {
+                result.Errors = result.Errors == null ? errors : result.Errors.Concat(errors);
+            }
+
+            if (result.Errors == null || !result.Errors.Any()
+                || options.RetryMode != RetryMode.OnErrorKeepDuplicates
+                && options.RetryMode != RetryMode.OnFatalKeepDuplicates) return result;
+
+            var duplicateErrors = (result.Errors ?? Enumerable.Empty<CogniteError>()).Where(err =>
+                err.Resource == ResourceType.ExternalId
+                && err.Type == ErrorType.ItemExists)
+                .ToList();
+
+            var duplicatedIds = new HashSet<InstanceIdentifier>();
+            if (duplicateErrors.Any())
+            {
+                foreach (var error in duplicateErrors)
+                {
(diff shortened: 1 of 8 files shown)


Toshad:
> ❌ Patch coverage is `51.82724%` with `145 lines` in your changes missing coverage. Please review.

There are subsequent PRs planned which will test some of this code too.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,364 @@
...
+                && err.Type == ErrorType.ItemExists)
+                .ToList();
+
+            var duplicatedIds = new HashSet<InstanceIdentifier>();
+            if (duplicateErrors.Any())
+            {
+                foreach (var error in duplicateErrors)
+                {
+                    if (error.Values == null || !error.Values.Any()) continue;
+                    foreach (var idt in error.Values) duplicatedIds.Add(idt.InstanceId);
```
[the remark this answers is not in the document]
Toshad: Only dm models here, not classic, so shouldn't be a problem.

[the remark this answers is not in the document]
Toshad: This is false anyway, instanceId is not optional, will be created at initialization.

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,6 +214,133 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+            SanitationMode sanitationMode,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            _logger.LogInformation("Getting or creating {Number} state sets in CDF", instanceIds.Count());
+            return await _client.Beta.StateSets.GetOrCreateStateSetsAsync(
+                instanceIds,
+                buildStateSets,
+                new BetaResourceParams(Chunking.Instances, Throttling.Instances, retryMode, sanitationMode),
+                token).ConfigureAwait(false);
+        }
```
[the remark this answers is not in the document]
Toshad: This is fine, count() on list/array is O(1), and you can only create 1k state sets max.

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,6 +214,133 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+            SanitationMode sanitationMode,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            if (stateSets == null) throw new ArgumentNullException(nameof(stateSets));
+            _logger.LogInformation("Ensuring that {Number} state sets exist in CDF", stateSets.Count());
+            return await _client.Beta.StateSets.EnsureStateSetsExistAsync(
+                stateSets,
+                new BetaResourceParams(Chunking.Instances, Throttling.Instances, retryMode, sanitationMode),
+                token).ConfigureAwait(false);
+        }
```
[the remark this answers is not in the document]
Toshad: This is fine, count() on list/array is O(1), and you can only create 1k state sets max.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,387 @@
...
+    {
+        private static ILogger _logger = new NullLogger<Client>();
+
+        internal static void SetLogger(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public static Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateAsync<T>(
+            ViewIdentifier view,
```
nithinb, line 62: No XML docs for any of the methods in this file?

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,6 +214,133 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+        /// <param name="token">Cancellation token</param>
+        /// <returns>A <see cref="CogniteResult{TResult, TError}"/> containing errors that occured and a list of the created and found state sets</returns>
+        public async Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateStateSetsAsync<T>(
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, Task<IEnumerable<SourcedNodeWrite<T>>>> buildStateSets,
+            RetryMode retryMode,
+            SanitationMode sanitationMode,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            _logger.LogInformation("Getting or creating {Number} state sets in CDF", instanceIds.Count());
```
nithinb:
```suggestion
            _logger.LogInformation("Getting or creating {Number} state sets in CDF", instanceIds?.Count());
```

nithinb: I think in general its safer to do this across the file?

nithinb (changes requested):
So had two comments, but more than that, considering the limit is 500, can you split this PR up? 

something like this? 

PR 1: Shared core + GetOrCreate
PR 2: Upsert + public wrappers
PR 3: Destination surface + integration

Toshad:
> So had two comments, but more than that, considering the limit is 500, can you split this PR up?
> 
> something like this?
> 
> PR 1: Shared core + GetOrCreate PR 2: Upsert + public wrappers PR 3: Destination surface + integration

Isn't the 500 line limit a soft limit?

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+            BetaResourceParams options,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            return BetaResourceExtensions.GetOrCreateAsync(
+                StateSetsResource.View,
+                (ids, tok) => stateSets.RetrieveAsync<T>(ids, tok),
+                (items, tok) => stateSets.UpsertAsync(items, null, tok),
+                DataModelSanitation.CleanInstanceRequest,
+                instanceIds, buildStateSets, options, token);
+        }
```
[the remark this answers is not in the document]
Toshad: Unnecessary as it is handled downstream. And we don't want to throw in those cases, just return.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+            CancellationToken token) where T : CogniteStateSet
+        {
+            if (options == null) throw new ArgumentNullException(nameof(options));
+            return BetaResourceExtensions.GetOrCreateAsync(
+                StateSetsResource.View,
+                (ids, tok) => stateSets.RetrieveAsync<T>(ids, tok),
+                (items, tok) => stateSets.UpsertAsync(items, null, tok),
+                DataModelSanitation.CleanInstanceRequest,
+                instanceIds, buildStateSets, options, token);
+        }
```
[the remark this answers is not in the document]
Toshad: Unnecessary as it is handled downstream. And we don't want to throw in those cases, just return.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,249 @@
...
+            RetrieveInstancesFunc<T> retrieve,
+            UpsertInstancesFunc<T> upsert,
+            SanitizeInstancesFunc<T> sanitize,
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, Task<IEnumerable<SourcedNodeWrite<T>>>> buildItems,
+            BetaResourceParams options,
+            CancellationToken token)
+        {
+            var chunks = instanceIds
```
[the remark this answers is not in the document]
Toshad: This is sematically wrong, in a lot of these cases, nulls are ok(e.g. santize can be null if SantiationMode is None). These params are best handled appropriately when needed.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,249 @@
...
+                        instanceIds.Select(x => new InstanceIdentifierWithType(InstanceType.node, x)), token).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    var err = ResultHandlers.ParseSimpleError<SourcedNodeWrite<T>>(ex, instanceIds?.Select(x => Identity.Create(x)), null);
+                    return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(new[] { err }, null);
+                }
+            }
+
+            var missing = instanceIds.Except(found.Select(ts => new InstanceIdentifier(ts.Space, ts.ExternalId))).ToList();
```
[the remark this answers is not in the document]
Toshad: Impossible unless the retrieve function is written wrongly, but will add the check to be careful.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            var duplicatedIds = new HashSet<InstanceIdentifier>();
+            if (duplicateErrors.Any())
+            {
+                foreach (var error in duplicateErrors)
+                {
+                    if (error.Values == null || !error.Values.Any()) continue;
+                    foreach (var idt in error.Values) duplicatedIds.Add(idt.InstanceId);
+                }
+            }
```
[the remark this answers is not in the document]
Toshad: State sets will only exist on CDM, and hence this should never be null, but will add a check.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="options">Chunking, throttling, retry, and sanitation options</param>
+        /// <param name="token">Cancellation token</param>
+        /// <returns>A <see cref="CogniteResult{TResult, TError}"/> containing errors that occured and a list of the created and found state sets</returns>
+        public static Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateStateSetsAsync<T>(
+            this StateSetsResource stateSets,
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, IEnumerable<SourcedNodeWrite<T>>> buildStateSets,
+            BetaResourceParams options,
+            CancellationToken token) where T : CogniteStateSet
+        {
```
vikramlc-cognite, line 37: Question: Why is the options guard not used here like here: https://github.com/cognitedata/dotnet-extractor-utils/pull/674/changes#diff-d83f96433d0c92b8c3638fefa9921b78951172f986c940126ed102e007076f69R65 ?

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
vikramlc-cognite, line 1: Question: Do you think we need more loggers in the file?

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
nithinb: I agree

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
Toshad: No, this is a generic writer. Logging should be done by callers which will have specific context.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
Toshad: Done.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceParams.cs
```diff
@@ -0,0 +1,45 @@
+namespace Cognite.Extensions.DataModels.CogniteExtractorExtensions
+{
+    /// <summary>
+    /// Bundles the chunking, throttling, retry, and sanitation parameters shared by the
+    /// beta CDM state-set and time-series extension methods (<see cref="BetaStateSetsExtensions"/>,
+    /// <see cref="BetaTimeSeriesExtensions"/>).
```
vikramlc-cognite: nit: **BetaTimeSeriesExtensions** -> This ref does not exist in this PR.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,249 @@
...
+        {
+            Task<IEnumerable<SourcedNodeWrite<T>>> AsyncBuildItems(IEnumerable<InstanceIdentifier> ids)
+            {
+                return Task.FromResult(buildItems(ids));
+            }
+            return GetOrCreateAsync(view, retrieve, upsert, sanitize, instanceIds, AsyncBuildItems, options, token);
+        }
+
+        /// <summary>
+        /// Retrieves instances by id for the beta CDM resource wrapped by a <see cref="BetaResourceExtensions"/> method. If any instances are missing, they will be created using the provided <see cref="buildItems"/> function.
```
vikramlc-cognite: nit: `<see cref="buildItems">` that should be `<paramref>`

nithinb (commented):
I agree with Vikram's comment, other than that, it looks good to me

vikramlc-cognite:
@Toshad can we add more unit tests for the file added/modified in this PR ? There are many logic added in many parts of the code which should be accompanied with unit tests.

Toshad:
> @Toshad can we add more unit tests for the file added/modified in this PR ? There are many logic added in many parts of the code which should be accompanied with unit tests.

Firstly, the PR is already too large, so I'll add UTs in a later one.
Secondly, This PR adds appropriate wrappers for CRUD operations on state sets, which is much better tested in Integration test. In unit test, we can only test some of the error handling. LMK if there's any other UTs you were thinking of.

vikramlc-cognite:
> Firstly, the PR is already too large, so I'll add UTs in a later one.
> Secondly, This PR adds appropriate wrappers for CRUD operations on state sets, which is much better tested in Integration test. In unit test, we can only test some of the error handling. LMK if there's any other UTs you were thinking of.

1. I am okay with adding UTs in a later PR but do mention it in the PR description and maybe open a draft PR and add the link here else risk review folks might point to it.
2. Agreed for the CRUD happy path, integration tests works fine. For unit tests, we can add for scenarios like Duplicate/backoff-retry logic, forcing real CDF to return a specific error type (fatal vs. non-fatal vs. none) on demand isn't something an integration test control, etc. I think claude can give a good list as well. This can definitely wait and be added in a later PR.

vikramlc-cognite (dismissed):
Approved

nithinb (approved):
LGTM

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaResourceExtensions.cs
```diff
@@ -0,0 +1,250 @@
...
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, IEnumerable<SourcedNodeWrite<T>>> buildItems,
+            BetaResourceParams options,
+            CancellationToken token)
+        {
+            Task<IEnumerable<SourcedNodeWrite<T>>> AsyncBuildItems(IEnumerable<InstanceIdentifier> ids)
+            {
+                return Task.FromResult(buildItems(ids));
+            }
+            return GetOrCreateAsync(view, retrieve, upsert, sanitize, instanceIds, AsyncBuildItems, options, token);
```
nithinb, line 75: I missed this in the previous review, this does not have an XML doc

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
nithinb: I don't think you addressed thsi comment?

nithinb (changes requested):
There are two changes I believe is worth doing as part of this PR.

Cognite.Extensions/DataModels/CogniteExtractorExtensions/BetaStateSetsExtensions.cs
Toshad: I think it gotapplied in another branch. Fixed it now.

nithinb (approved):
🦄 

The risk of not having all the unit tests is something we are accepting since it has all the integration tests. And since this is released as part of versioning this will not get released without the appropriate unit tests.

PR #688, opened 2026-07-27
Add CogniteDestinationWithIDM state-set convenience methods

Toshad:
Adds GetOrCreateStateSetsAsync, EnsureStateSetsExistAsync,
GetStateSetsByIdsIgnoreErrors, and UpsertStateSetsAsync to
CogniteDestinationWithIDM, forwarding to the BetaStateSetsExtensions
layer with chunking/throttling derived from CogniteConfig. Builds on
the beta CDM state-set core added previously.

The accompanying test is a smoke test only: these are thin forwarding
wrappers with no logic of their own, and the retry/sanitation/duplicate
behavior they delegate to is already covered by the dedicated tests
against BetaStateSetsExtensions directly.


ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,13 +214,143 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+        {
+            if (instanceIds == null) return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(null, null);
+            if (buildStateSets == null) throw new ArgumentNullException(nameof(buildStateSets));
+            _logger.LogInformation("Getting or creating {Number} state sets in CDF", instanceIds.Count());
+            return await _client.Beta.StateSets.GetOrCreateStateSetsAsync(
+                instanceIds,
+                buildStateSets,
+                new BetaResourceParams(Chunking.Instances, Throttling.Instances, retryMode, sanitationMode),
+                token).ConfigureAwait(false);
+        }
```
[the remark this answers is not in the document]
Toshad: This is OK for 2 reasons:
1. In-memory pass through a list is still fast, even for 1M elements.
2. The callers already pass a list or some other ICollection.

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,13 +214,143 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+        /// <returns>A <see cref="CogniteResult{TResult, TError}"/> containing errors that occured and a list of the created and found state sets</returns>
+        public async Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateStateSetsAsync<T>(
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, Task<IEnumerable<SourcedNodeWrite<T>>>> buildStateSets,
+            RetryMode retryMode,
+            SanitationMode sanitationMode,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            if (instanceIds == null) return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(null, null);
+            if (buildStateSets == null) throw new ArgumentNullException(nameof(buildStateSets));
```
vikramlc-cognite, line 267: Question: Should we check for buildStateSets first and then instanceIds as one throws exception and one returns empty ?

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,13 +214,143 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
...
+        /// <returns>A <see cref="CogniteResult{TResult, TError}"/> containing errors that occurred and a list of the created and found state sets</returns>
+        public async Task<CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>> GetOrCreateStateSetsAsync<T>(
+            IEnumerable<InstanceIdentifier> instanceIds,
+            Func<IEnumerable<InstanceIdentifier>, IEnumerable<SourcedNodeWrite<T>>> buildStateSets,
+            RetryMode retryMode,
+            SanitationMode sanitationMode,
+            CancellationToken token) where T : CogniteStateSet
+        {
+            if (instanceIds == null) return new CogniteResult<SourcedNode<T>, SourcedNodeWrite<T>>(null, null);
+            if (buildStateSets == null) throw new ArgumentNullException(nameof(buildStateSets));
```
vikramlc-cognite, line 238: Question: Should we check for buildStateSets first and then instanceIds as one throws exception and one returns empty ?

nithinb (changes requested):
The change itself looks good. I think we just need some tests on the following:

* one happy-path test calling CogniteDestinationWithIDM.EnsureStateSetsExistAsync directly

* one happy-path test for the sync-builder GetOrCreateStateSetsAsync overload.

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
Toshad: The correctness is debatable, but I prefer the current one because buildStateSets is necessarily something that applies to each member of instance ids, and this makes the method less likely to throw.

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
Toshad: ^

vikramlc-cognite (dismissed):
Approved

nithinb (changes requested):
So 2 things:

* You are making a release on this branch, but would it make more sense to make the release after timeseries_datapoints is merged? 

* StateSet param has no integration test as far as I can see.

Toshad:
> So 2 things:
> 
> * You are making a release on this branch, but would it make more sense to make the release after timeseries_datapoints is merged?

No, since the state sets methods are self-contained, this is more appropriate.

> * StateSet param has no integration test as far as I can see.

What do you mean by `StateSet param`?

ExtractorUtils/Cognite/CogniteDestinationWithIDM.cs
```diff
@@ -213,13 +214,143 @@ public async Task<CogniteResult<SlimInstance, SourcedNodeWrite<T>>> UpsertTimeSe
         }
         #endregion
 
+        #region state sets
+        /// <summary>
+        /// Ensures the the state sets with the provided <paramref name="instanceIds"/> exist in CDF.
```
nithinb, line 219: nit and non-blocking: There are multiple locations with the phrase being 'the the' 
Can you fix that in a subsequent PR?

nithinb (approved):
risk review ok

